Issue #931, opened 2014-12-10
search's cleared event is fired when field is empty


interactivellama:
You can see it in Matt's JSbin http://jsbin.com/nupukiyake/

To duplicate: Search for text, then backspace until the field is clear.

The repeater reloads when the value of the textfield is blank, because the `cleared`event is being fired. 

Cleared should only be fired if enter is hit with empty value or the 'x' button is hit. 

Tabbing out and tabbing back in should not fire `cleared` either.

PR #932, opened 2014-12-10
search's cleared event is fired when field is empty #931

BenjaminNeilDavis:
- added an option to clear when input is empty (by default set to false)
- worked with the search / close icon to make more sense
- When already searched a value the icon turns to a x. The change i implemented is if the input value changes the icon turns back to a search icon. Before it would stay an x and would have to click enter to search a different value.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -228,6 +228,16 @@ module.exports = function(grunt) {
 				}
 			}
 		},
+		replace: {
+			dist: {
+				src: ['DETAILS.md', 'README.md'],
+				overwrite: true,                 // overwrite matched source files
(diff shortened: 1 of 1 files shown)


interactivellama:
That third bullet is a great UX find.

interactivellama:
After I clear the field, the icon goes back to search, but when I tab back into the input field. The clear icons appears again even though the field is blank. (last frame of screen capture below)

swilliamset:
@BenjaminNeilDavis please update your branch so the PR can be merged.

BenjaminNeilDavis:
Working on the fix now...great catch

Issue #946, opened 2014-12-17
Pillbox edited event is not preserving existing value attribute for tag


tonyzampogna:
When editing an existing tag with a value attribute (data-value='something'), that value is not sent along in the arguments when the edited.fu.pillbox is triggered.

interactivellama:
I'm going to need some more information. I am getting `{value: "test", text: "test"}` from `edited.fu.pillbox`.

swilliamset:
Problem is the markup isn't getting updated as explained here https://github.com/ExactTarget/fuelux/issues/960

PR #947, opened 2014-12-17
Pillbox edited event is not preserving existing value attribute for tag #946

tonyzampogna:
This is a possible patch for issue #946.

When editing an existing tag with a value attribute (data-value='something'), that value is not sent along in the arguments when the edited.fu.pillbox is triggered.

### DIFF
--- a/js/pillbox.js
+++ b/js/pillbox.js
@@ -349,6 +349,11 @@
 
 			if (this.acceptKeyCodes[e.keyCode]) {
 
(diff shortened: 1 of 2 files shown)


swilliamset:
The solution provided passes along the out of date value from $.data('value') and does not speak to the underlying problem of the $.data('value') not being update as mentioned in https://github.com/ExactTarget/fuelux/issues/960. Closing PR but we will certainly continuing digging into syncing these values.

PR #961, opened 2015-01-02
moment deprecated "lang" for "locale"

swilliamset:


interactivellama:
Fuel UX now optionally requires Moment 2.8 or greater due to locale method.

interactivellama:
Now considering supporting < moment 2.8 additionally.

swilliamset:
see https://github.com/ExactTarget/fuelux/pull/962

PR #963, opened 2015-01-05
issue-931-search-cleared-event

BenjaminNeilDavis:
fixing backspace to empty icon change


interactivellama:
Works great. Thanks for working through this!

PR #971, opened 2015-01-09
Repeater Different Columns Bugfix

### LINKED ISSUE #970: Additional repeater columns check to see if columns have changed
The function that checks if columns are different neglects to check column array lengths, resulting in odd behavior. This needs to be fixed so that dynamic columns are truly possible

kevinparkerson:
Fixes #970 by also checking incoming column array lengths to determine whether a change has occurred

### DIFF
--- a/dev/dev.html
+++ b/dev/dev.html
@@ -38,31 +38,191 @@
 				}
 			});
 		})();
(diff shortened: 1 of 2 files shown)


interactivellama:
Thanks for the quick turn around.

PR #978, opened 2015-01-15
(gh-pages) :lipstick: :bug: :arrow_up: :fire: Merge Form Builder into ge...

cormacmccarthy:
...tfuelux.com. Refactor codebase to make things more legible. Make textarea fields actually. Update Fuel UX dependencies to ~3.5.0


futuremint:
Is there a build step in here somewhere where all of the require.js dependencies get optimized into a single request so that require.js doesn't make a ton of HTTP requests for the assets.

cormacmccarthy:
Yeah, so... That would make sense... but... not yet. If that's a deal breaker, we can wait to merge until I have that done. Otherwise, I can try and have that done by EOW and will issue a new PR.

Issue #980, opened 2015-01-16
Selectlist: Auto sizing not working again


mbeard:
A couple additional issues i'm seeing:
1. If the selectlist is hidden when initialized is fired, it's setting the width of the menu and button to 0px.
![screenshot 2015-01-16 13 59 55](https://cloud.githubusercontent.com/assets/467352/5782365/69d5e06a-9d88-11e4-9e28-cb8497b992d7.png)

2. The logic to set the width isn't accounting for the width of the caret for the dropdown
![screenshot 2015-01-16 14 00 55](https://cloud.githubusercontent.com/assets/467352/5782368/74f24858-9d88-11e4-9f45-59e99a16f7c0.png)

3. The button text is centered and not left aligned

4. Need to consider the headers as well, when looking for the longest string.  For example, a menu like this:
![screenshot 2015-01-16 14 01 59](https://cloud.githubusercontent.com/assets/467352/5782370/7d5c787e-9d88-11e4-8de0-041e56a343c8.png)

cormacmccarthy:
@BenjaminNeilDavis has a fix which he will be checking in tonight.

PR #982, opened 2015-01-17
Selectlist: Auto sizing not working again #980

BenjaminNeilDavis:
Selectlist: Auto sizing not working again #980


BenjaminNeilDavis:
@mbeard Reason I did that is because I have to create the elements and have them display the same way they normally would to figure out the sizes.

Issue #987, opened 2015-01-19
Use .hidden on checkbox and radio toggle elements


interactivellama:
`.hide` is deprecated and `.hidden` should be used. radio and checkbox JS files need to updated. Should remove `.hide` and `.hidden` and then only add `.hidden` back.

See http://getbootstrap.com/css/#helper-classes-show-hide

PR #999, opened 2015-01-22
Use .hidden on checkbox and radio toggle elements #987

BenjaminNeilDavis:
- Fixes issue #987
- checkbox and radio remove both .hide and .hidden but only add .hidden.

### DIFF
--- a/js/checkbox.js
+++ b/js/checkbox.js
@@ -129,10 +129,10 @@
 		toggleContainer: function(){
 			if( Boolean( this.$toggleContainer ) ) {
 				if( this.state.checked ) {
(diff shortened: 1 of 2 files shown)


cormacmccarthy:
From Bootstrap's docs:

> .hide is available, but it does not always affect screen readers and is deprecated as of v3.0.1. Use .hidden or .sr-only instead.

This change updates Fuel UX to be compatible with this standard.

cormacmccarthy:
More is needed to bring us into compliance here, but I've created Issue #1002 to address these other occurrences.

PR #1005, opened 2015-01-22
wizard: Dont steal focus when chaning steps.

sergio91pt:
Allowing to change or remove focus in the actionclicked callback.
This is specially important on form steps without required fields.

### DIFF
--- a/js/wizard.js
+++ b/js/wizard.js
@@ -294,14 +294,15 @@
 				this.setState();
 			}
 
-			// return focus to control after selecting an option
-			if( this.$prevBtn.is(':disabled') ) {
-				this.$nextBtn.focus();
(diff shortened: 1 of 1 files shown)


cormacmccarthy:
@sergio91pt good catch! I like the idea of trying to allow for more intelligent decision making around where to put `focus()` to make it the most useful for the user. This is a good first step, and we are going to explore taking this a little further. Rather than _ever_ focusing on the prev or next button, we may want to focus, by default, on the first form field within the wizard itself.

I'm going to leave this open until we have a chance for a little more discussion/exploration here.

cormacmccarthy:
This PR is included in PR #1006 

Thanks!

PR #1006, opened 2015-01-23
1005   correctly apply focus on step change in wizard

cormacmccarthy:
:bug: Avoid clobbering programmatically set focus
:new: If possible, focus on the first applicable form field instead of on next/prev button if focus has not been set programmatically
:lipstick::racehorse: clean up code a little to do less js processing, improve readability, and better conform to JS standards

### DIFF
--- a/js/wizard.js
+++ b/js/wizard.js
@@ -8,7 +8,7 @@
 
 // -- BEGIN UMD WRAPPER PREFACE --
 
-// For more information on UMD visit: 
+// For more information on UMD visit:
 // https://github.com/umdjs/umd/blob/master/jqueryPlugin.js
 
 (function (factory) {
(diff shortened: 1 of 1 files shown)


js/wizard.js
```diff
@@ -281,51 +275,54 @@
...
 			}
 
-			// return focus to control after selecting an option
-			if( this.$nextBtn.is(':disabled') ) {
-				this.$prevBtn.focus();
-			}
-			else {
-				this.$nextBtn.focus();
+			// only set focus if focus is still on the $nextBtn (avoid stomping on a focus set programmatically in actionclicked callback)
+			if( this.$nextBtn.is(":focus") ) {
```
kevinparkerson: This is minor and nit-picky, but could you change this line to ':focus' instead? It'd be more consistent with how strings are defined elsewhere in the file.

cormacmccarthy: sure thing!

kevinparkerson:
This looks good, although the Travis build is failing. Re-running, might need to check if something else changed that is causing this.

PR #1014, opened 2015-01-29
Fix the basic recursion error using setTimeout

### LINKED ISSUE #774: Repeater - too much recursion
Currently the Repeater widget is simply broken in Firefox. Even on the official example site (http://getfuelux.com/javascript.html#repeater) the Repeater stops working, when I set the number of rows to 50 or more. The error that the browser reports is "too much recursion", and it dies after a few hundred cell drawing. We've tried to use it for a very large table (30+ columns), in that case it can't render 10 rows in Firefox.

tweettypography:
Using setTimeout breaks up the call stack. This fixes #774

### DIFF
--- a/js/repeater.js
+++ b/js/repeater.js
@@ -644,10 +644,10 @@
 					}else{
 						callbacks[stage](null);
 					}
-				};
+				}
 				
 				proceed('before', args);
-			};
+			}
 
 			if(renderer.repeat){
 				repeat = renderer.repeat.split('.');
--- a/js/repeater.js
+++ b/js/repeater.js
@@ -547,6 +547,7 @@
 			var self = this;
 			var skipNested = false;
 			var index = 0;
+			var skip = 0;
 			var repeat, subset, i, l;
 
 			function loopSubset(){
@@ -603,16 +604,20 @@
 					after: function(resp){
 						var cont;
 						var loopNested = function(cont, index){
-							setTimeout(function(){
-								self.runRenderer(cont, renderer.nested[index], data, function(){
-									index++;
-									if(index<renderer.nested.length){
-										loopNested(cont, index);
-									}else{
+							self.runRenderer(cont, renderer.nested[index], data, function(){
+								index++;
+								if(index<renderer.nested.length){
+									loopNested(cont, index);
+								}else{
+									if (++skip % 50 === 0) {
+										setTimeout(function(){
+											proceed('complete', args);
+										}, 1);
+									} else {
 										proceed('complete', args);
 									}
-								});
-							}, 0);
+								}
+							});
 						};
 
 						if(resp && resp.skipNested===true){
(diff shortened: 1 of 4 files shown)


js/repeater.js
```diff
@@ -638,7 +645,7 @@
 						callbacks[stage](null);
 					}
 				};
```
kevinparkerson: Grunt is complaining about these unnecessary semicolons. There's another on line 650. Symptom of converting from the variable-declared functions I imagine.

js/repeater.js
```diff
@@ -546,9 +546,21 @@
 		runRenderer: function(container, renderer, data, callback){
 			var self = this;
 			var skipNested = false;
+			var index = 0;
 			var repeat, subset, i, l;
 
-			var loopSubset = function(index){
+			function loopSubset(){
```
kevinparkerson: while I prefer declaring functions this way over creating a variable, in the past jsHint would yell about that... however, it does not seem to be complaining currently. Did we alter the jsHint settings? If so and we're cool with that, great! :+1:

kevinparkerson:
Tested this out, it allows for HUGE number of rows to be used (I tried 10,000) which is pretty awesome. However, speed is significantly slower, even at say 100 rows. Not sure if there's anything that can be done about that.

js/repeater.js
tweettypography: Good point! Just a typo

js/repeater.js
tweettypography: I changed it because I read some studies showing that it was potentially *slightly* faster and with the amount of function calls happening in this recursive approach it might be worth it

tweettypography:
I might be able to improve the speed a little without sacrificing too much. Let me take a look. For an average number of rows I felt like the difference wasn't very noticeable, and for large numbers the grid never worked in the first place, but I'll see if I can tweak slightly

tweettypography:
Updated. 10,000 rows might still be a bit slow but more typical numbers shouldn't suffer the same fate

js/repeater.js
```diff
@@ -603,7 +609,13 @@
 								if(index<renderer.nested.length){
 									loopNested(cont, index);
 								}else{
-									proceed('complete', args);
+									if (++skip % 50 === 0) {
```
kevinparkerson: We should make this value a configurable option so that this can be optimized best for individual repeater performance. Perhaps a name like `recursionLimit' or something along those lines

futuremint: Lets wait on that to see if this can be done without recursion altogether.

tweettypography: The problem is that we have an API that is based on callbacks for the render / sub-render steps. I tried a lot of different solutions but kept running into the same exponential growth because of that. Prior to realizing I could use setTimeout as a sort of trick/hack I was feeling like we were going to have to deprecate the API and start over (which might still be worthwhile but is obviously a breaking change)

futuremint:
Merging this interim solution for now.

PR #1020, opened 2015-02-02
move label away from float and space caret in right side padding (fix #1021)

### LINKED ISSUE #1021: Selectlist caret squishes: needs more stable spacing
screen shot 2015-02-02 at 4 39 47 pm

swilliamset:


less/selectlist.less
```diff
@@ -13,15 +17,14 @@
 		min-width: inherit;
 	}
 	.btn.dropdown-toggle {
-		white-space: normal;
+		white-space: nowrap;
```
futuremint: This is already set on .btn.  Might as well just remove this line completely if you're setting it back to nowrap.

swilliamset:
fixes #1021

PR #1041, opened 2015-02-06
Repeater plugin api refactor

### LINKED ISSUE #1022: Repeater Plugin API too complex
The current repeater plugin API is too complex. Aside from being relatively difficult to work with for more advanced plugins, it also causes problems creating an efficient repeater as the call stack is exceeded due to the heavy recursion needed to implement it. A simplified plugin API has been drafted and should be implemented ASAP:

```js
{
	cleared: function(helpers){},					//called when the repeater is cleared
    dataOptions: function(helpers){},				//plugin uses this to add additional dataOptions prior to calling the dataSource
    initialize: function(helpers, callback){},		//called once upon repeater initialization. callback must be called to continue
    resize: function(helpers){},					//called when the repeater is resized (happens when window is resized or resize is called via API)
    selected: function(helpers){},					//called when the view is selected, before rendering
    
    before: function(helpers){},					//called prior to rendering items, if render function not defined
    renderItem: function(helpers){},				//called per item, if render function not defined. iterates on defined repeat namespace
    after: function(helpers){},						//called after rendering items, if render function not defined
    repeat: 'data.items',							//specifies namespace to repeat on. default is "data.items"
    
    render: function(helpers, callback){}			//called instead of before, renderItems, and after methods if defined. callback must be called to continue
}

// OLD version for comparison

//views object contains keyed list of view plugins, each an object with following optional parameters:
        //{
            //cleared: function(helpers, callback){},
            //dataOptions: function(helpers, callback){},
            //initialize: function(helpers, callback){},
            //selected: function(helpers, callback){},
            //resize: function(helpers, callback){},
            //renderer: {}
        //}
            //renderer object contains following optional parameters:
                //{
                    //before: function(helpers, callback){},
                    //after: function(helpers, callback){},
                    //complete: function(helpers, callback){},
                    //repeat: 'parameter.subparameter.etc',
                    //render: function(helpers, callback){},
                    //nested: [ *array of renderer objects* ]
                //}

                //helpers object structure:
                    //{
                        //container: jQuery object,    (current renderer parent)
                        //data: {...}, (data returned from dataSource)
                        //index: int, (only there if repeat was set. current item index)
                        //item: str or jQuery object, (only there if rendered function returned item)
                        //subset: {}, (only there if repeat was set. subset of data being repeated on)
                    //}
```

kevinparkerson:
Fixes #1022 by implementing new repeater plugin API. All other APIs remain the same.

### DIFF
--- a/dev/dev.html
+++ b/dev/dev.html
@@ -92,7 +92,7 @@
 					// from data.js
 					resp.items.push(data.repeater.listData[j]);
 					j++;
-					if(j > data.repeater.listData.length){
+					if(j >= data.repeater.listData.length){
 						j = 0;
 					}
 				}
@@ -173,7 +173,7 @@
 						list_highlightSortedColumn: true
 					},
 					'thumbnail.view3': {
-						thumbnail_alignment: 'center',
+						thumbnail_alignment: 'none',
 						dataSource: function(options, callback){
 							thumbnail(options, callback);
 						},
--- a/dev/dev.html
+++ b/dev/dev.html
@@ -15,298 +15,6 @@
 	<script type="text/javascript">if(window.console && window.console.clear){ window.console.clear(); }</script>
 
 	<script src="../bower_components/requirejs/require.js" type="text/javascript"></script>
-	<style>
-		#triggerModal .modal-content {
-			/* new custom width */
-			width: 1100px;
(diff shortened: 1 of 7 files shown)


interactivellama:
NOTE: `rendered` and `loaded` events have changed a little on when they get triggered.

PR #1047, opened 2015-02-06
Gh1044   restructure less build to allow ampersand to not have fuelux in parents

chriscorwin:
## Purpose
The proposed change would allow us to begin to use the `&` parent selector that LESS provides more effectively.

## Overview
The less build should be able to be slightly restructured so that the `.fuelux` wrapper class name is not in the parent-chain (`&`) during the main LESS compilation, and instead gets tacked onto the front of the rules after the fact. 

The CSS file ultimately output ought not be any different after making this change, since the `&` selector currently goes under-utilized (due, of course, to the quirks of our current process).

## Simple example 

From the LESS documentation on [Changing Selector Order](http://lesscss.org/features/#parent-selectors-feature-changing-selector-order).

Given the following code:

```
.header {
  .menu {
    border-radius: 5px;
    .no-borderradius & {
      background-image: url('images/button-background.png');
    }
  }
}
```

...the way we are currently building the LESS would render the .css as:

```
.fuelux .header .menu {
  border-radius: 5px;
}
.no-borderradius .fuelux .header .menu {
  background-image: url('images/button-background.png');
}
```

Note the stray `.fuelux` that is inserted _after_ the `.no-borderradius` selector, which is _not_ what the developer intended.

The restructure would output as:

```
.fuelux .header .menu {
  border-radius: 5px;
}
.fuelux .no-borderradius .header .menu {
  background-image: url('images/button-background.png');
}
```

## The Actual Restructure Proposal

Change the [fuelux.less](https://github.com/ExactTarget/fuelux/blob/master/less/fuelux.less) file so that it does the its initial compilation _not inside the `.fuelux` wrapper_.

It would output a temporary "non-namespaced" file, perhaps located at `/less/temp.less`.

At that point we would process a new LESS file which imports the temporary file we created -- this time _inside_ the wrapper -- tacking the `.fuelux` "namespace" onto the rules.

```
@charset "UTF-8";

.fuelux {
	@import (less) "temp.less";
}
```

The grunt task doing this would output the resulting .css into it's final home at `/dist/css/fuelux.css`, and then delete the temporary file.

The temporary file should _also_ be added to `.gitignore` to ensure that a failed LESS build does not result in the temp file not being deleted and it being inadvertently checked into the repository.


.gitignore
```diff
@@ -19,4 +19,7 @@ SAUCE_API_KEY.yml
 .idea
 
 # node_module cache for TravisCI, etc.
-node_modules.tar.gz
\ No newline at end of file
+node_modules.tar.gz
+
+# Temp file created int he LESS build
```
cormacmccarthy: typo "int he"

chriscorwin: Fixed, thanks.

Gruntfile.js
```diff
@@ -413,12 +426,22 @@ module.exports = function(grunt) {
 		grunt.task.run(['bump-only:' + versionReleaseType, 'dist', 'replace:readme']);
 	});
 
+
+	/* -------------
+		SERVEFAST
+	------------- */
+	grunt.registerTask('servefast', 'Serve the files with no "dist" build or tests. Optional --no-less to also disabled compiling less into css.', function() {
```
cormacmccarthy: disabled -> disable

cormacmccarthy: several trivial jshint errors here too actually, I'm fixing so I can merge the PR.

chriscorwin: Thanks, I couldn't figure out on a Friday afternoon what the heck was going on with travis errors.

Nor am I sure why or how they were introduced.

I am gonna dig in and look at what happened, but... you didn't remove the enhancements to "servefast" did you?

Those kinda sneaked in this, but were the result of an afternoon playing with it with @interactivellama .

PR #1057, opened 2015-02-09
move/add id to input for better radio button handling (fix #1056)

### LINKED ISSUE #1056: form builder adds `id` to `label` instead of `input` for radio buttons
https://github.com/ExactTarget/fuelux/blob/gh-pages/assets/js/form-builder/templates/component/multipleradiosinline.html#L6

screen shot 2015-02-09 at 11 21 58 am

swilliamset:


assets/js/form-builder/templates/component/multipleradios.html
```diff
@@ -5,7 +5,7 @@
 		<% _.each(radios, function(radio, i){ %>
 			<div class="radio" data-initialize="radio" id="<%= name+'-'+i %>">
 				<label class="radio-custom" for="<%= name+'-'+i %>">
-					<input class="sr-only" name="<%= name+'-'+i %>" type="radio" value="<%= radio %>">
+					<input class="sr-only" id="<%= name+'-'+i %>" name="<%= name %>" type="radio" value="<%= radio %>">
```
cormacmccarthy: This puts duplicate ids on the page. You'll need to remove id from line 6.

cormacmccarthy:
This change makes me nervous. I'll need to test all the ramifications...

swilliamset:
Sorry about that. should be good now.

cormacmccarthy:
OK, looks like I was already protecting against the thing I was worried this would impact:

    		this.$radio = $(element).is('input[type="radio"]') ? $(element) : $(element).find('input[type="radio"]:first');

I'm _ok_ with making this change, but, if this is going to be the recommendation going forward, we need to update the documentation as well (the FormBuilder code matched the documentation with the exception that it added a "for" attribute where the docs did not have one)

swilliamset:
IMHO we should encourage good markup (using `for`). As for this bug if we can call it fixed lets do so, merge, and update the docs. New issue https://github.com/ExactTarget/fuelux/issues/1062.

PR #1058, opened 2015-02-09
Add destroy support for inline radio

interactivellama:
Currently `destroy()` searches for `.radio` to remove control, now it will remove label if `.radio` does not exist. Fixes. #1059

### DIFF
--- a/js/radio.js
+++ b/js/radio.js
@@ -35,11 +35,12 @@
 		this.$radio = $(element).is('input[type="radio"]') ? $(element) : $(element).find('input[type="radio"]:first');
 		this.$label = this.$radio.parent();
 		this.groupName = this.$radio.attr('name');
-		this.$parent = this.$label.parent('.radio');
+		this.$blockWrapper = this.$label.parent('.radio');	// only used if block radio control, otherwise radio is inline
+		this.isBlockWrapped = true;	// initialized as a block radio control
 		this.$toggleContainer = null;
 
-		if (this.$parent.length === 0) {
-			this.$parent = null;
+		if (this.$blockWrapper.length === 0) {
+			this.isBlockWrapped = false;
 		}
 
 		var toggleSelector = this.$radio.attr('data-toggle');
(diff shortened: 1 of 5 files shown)


js/radio.js
```diff
@@ -59,13 +59,20 @@
...
 			// [none]
 			// empty elements to return to original markup
 			// [none]
 			// return string of markup
-			return this.$parent[0].outerHTML;
+			if (this.$parent) {
+				this.$parent.remove();
+				return this.$parent[0].outerHTML;
+			}
+			else {
```
cormacmccarthy: please combine line 70 & 71, else should not start a new line.

js/radio.js
```diff
@@ -59,13 +59,20 @@
...
 
 		destroy: function () {
-			this.$parent.remove();
 			// remove any external bindings
 			// [none]
 			// empty elements to return to original markup
 			// [none]
 			// return string of markup
-			return this.$parent[0].outerHTML;
+			if (this.$parent) {
```
swilliamset: should be `if( this.$parent.length )`. Empty jquery selector is truthy.

cormacmccarthy: Well, actually we already do the check earlier in the file:

		if (this.$parent.length === 0) {
			this.$parent = null;
		}

cormacmccarthy:
We need some badges. You should get a badge for "remembered to think about unit tests" ;)

Here's a star: :star:

js/radio.js
swilliamset: oh i see it now. certainly could be adjusted. or add a `this.isWrapped` attribute to be initialized with `true` and set to false with 41-43

cormacmccarthy: Oh, good call! I like it. :+1:

js/radio.js
```diff
@@ -35,11 +35,13 @@
...
 		this.groupName = this.$radio.attr('name');
-		this.$parent = this.$label.parent('.radio');
+		this.$blockWrapper = this.$label.parent('.radio');	// only used if block radio control, otherwise radio is inline
+		this.isBlockWrapped = true;	// initialized as a block radio control
 		this.$toggleContainer = null;
 
-		if (this.$parent.length === 0) {
-			this.$parent = null;
+		if (this.$blockWrapper.length === 0) {
+			this.$blockWrapper = null;
```
cormacmccarthy: I actually don't think we need to set this to null. I think that we were setting $parent to null to make boolean checking easier. That was a bad practice. Let's just leave $blockWrapper alone here...

interactivellama: Yeah, now that we are using the boolean there's no need for it. It's legacy code from previous implementation.

cormacmccarthy:
I believe this affects checkbox too, no? Can you update checkbox as well please?

cormacmccarthy:
Semi-related checkbox issue #1066 being handled separately

PR #1073, opened 2015-02-10
Solution needed for selectlist [data-resize]=auto when `display: none`- fixes #1069

### LINKED ISSUE #1069: Solution needed for selectlist [data-resize]=auto when `display: none`
If the control is hidden when initiated, then the width is 0px.

screen shot 2015-02-10 at 12 15 02 pm

http://jsbin.com/nuxugocuwa

BenjaminNeilDavis:
- fixes #1069
- Solution needed for selectlist [data-resize]=auto when `display: none`
- added catch if width returns as 0 and displays selectlist as normal.
- different test examples http://jsbin.com/beyogiqizi/


less/misc.less
```diff
@@ -53,6 +53,9 @@ input, textarea {
 	visibility: hidden;
 	top: 0;
 	float: left;
+	.selectlist {
+		display: block !important;
```
swilliamset: uncomfortable with an `!important` here. is there a place we can do the following
* check if visible
* if hidden 
 * change visibility transparent
 * show
 * measure
 * reset

BenjaminNeilDavis: This is just for the dummy selectlist I create. The element I append to the body to get the width of the dropdown items then remove once I have a width.

swilliamset: lets provide a more explicit class name then to make it clear it's a dummy
`.dummy-selectlist`

cormacmccarthy: If "sizer" is good enough for the js variable name, why are we deviating from that in the CSS? Now it is inconsistent and _even more confusing_.

cormacmccarthy:
Are we able to test this in a unit test?

cormacmccarthy:
Your "hidden on load" test example is breaking:

image

less/misc.less
swilliamset: Sizer is fine. We can move back. I wasn't looking at the context of the css class.

On Feb 11, 2015 12:43 PM, Christopher McCulloh <notifications@github.com> wrote:

In less/misc.less<https://github.com/ExactTarget/fuelux/pull/1073#discussion_r24515227>:

> @@ -53,6 +53,9 @@ input, textarea {
>       visibility: hidden;
>       top: 0;
>       float: left;
> +     .selectlist {
> +             display: block !important;


If "sizer" is good enough for the js variable name, why are we deviating from that in the CSS? Now it is inconsistent and even more confusing.

BenjaminNeilDavis:
@cormacmccarthy Were you testing it in the selectlist test markup?

cormacmccarthy:
I clicked on the link you provided:

image

(http://jsbin.com/beyogiqizi/)

and it was borked.

BenjaminNeilDavis:
@cormacmccarthy I was just mainly providing the html for the tests (because dev file is now ignored). The js / css for fuelux is pointing to the CDN which doesnt have my changes yet. If you were to copy the html / js from the example and use it on the dev.html page you would be able to see the change.

cormacmccarthy:
Indeed! Thanks :)

image

cormacmccarthy:
Travis build failing because it isn't using phantomjs (I think). PR #1084 fixes this issue. Once that PR is merged (and this PR is updated with that) the travis build will pass.

cormacmccarthy:
I think a bad squash resulted in losing some code.

PR #1074, opened 2015-02-10
ensured the scheduler input checked attribute is being set, not just setting the active css class

### LINKED ISSUE #992: Scheduler reoccurance pattern loses weekday
Bug found here: http://stackoverflow.com/questions/27974320/fuelux-scheduler-loose-the-information-of-day-when-we-select-weekly-recurrence

Change [L510](https://github.com/ExactTarget/fuelux/blob/master/js/scheduler.js#L510) to something like the following to include `attr('checked', true)`: `item.find( 'input[data-value="' + temp[ i ] + '"]' ).attr('checked', true).parent().addClass( 'active' );`

Add additional unit tests to test multiple reoccurrence patterns.

mbeard:
fixes #992

### DIFF
--- a/js/scheduler.js
+++ b/js/scheduler.js
@@ -515,7 +515,7 @@
 						item.find('label').removeClass('active');
 						temp = recur.BYDAY.split(',');
 						for (i = 0, l = temp.length; i < l; i++) {
-							item.find('input[data-value="' + temp[i] + '"]').attr('checked',true).parent().addClass('active');
+							item.find('input[data-value="' + temp[i] + '"]').prop('checked',true).parent().addClass('active');
 						}
 					}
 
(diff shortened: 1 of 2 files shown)


interactivellama:
I wonder if that has anything to do with issue #992.

swilliamset:
@mbeard travis is failing on test failing

js/scheduler.js
```diff
@@ -515,7 +515,7 @@
 						item.find('label').removeClass('active');
 						temp = recur.BYDAY.split(',');
 						for (i = 0, l = temp.length; i < l; i++) {
-							item.find('input[data-value="' + temp[i] + '"]').parent().addClass('active');
+							item.find('input[data-value="' + temp[i] + '"]').attr('checked',true).parent().addClass('active');
```
cormacmccarthy: It doesn't look like your unit test is checking for the checked attr to be present and true, is this on purpose?

mbeard: I was intending to test getValue which in turn tests this indirectly.  I will create a separate unit test to explicitly test the `checked` attribute.

cormacmccarthy: I believe this should be `.prop('checked', true)` not `attr('checked', true)`. http://api.jquery.com/prop/

    the most important concept to remember about the checked attribute is that it does not correspond to the checked property. The attribute actually corresponds to the defaultChecked property and should be used only to set the initial value of the checkbox. The checked attribute value does not change with the state of the checkbox, while the checked property does.

cormacmccarthy:
OK, I can confirm that `.attr('checked')` is not getting set "under the covers":

image

But if we set it, it _will be_ set:

image

and furthermore, if we use attr() instead of prop(), despite what the jquery docs tell us, it appears to work better:

image

mbeard:
Why not rely on our defined API for interrogating the scheduler vs. supporting inspecting the classnames, attributes, etc. on the DOM elements of the scheduler?

The latter seems to be problematic and cause for future issues.

cormacmccarthy:
This has possible ramifications for anyone who was incorrectly accessing the value using `attr` instead of interrogating our API using `getValue` or `change` event (or, worst case, using `prop`)

interactivellama:
We have something in that essentially said we will legacy support our API and initial markup if we change it, custom DOM interrogating though may break, but I can't find it now.

[Here it is](http://getfuelux.com/javascript.html#overview-events). It's only about events though. 

```
 Listening for standard and Bootstrap supported events on an internal element is not tested, and therefore not recommended.
```

We could add a disclaimer for markup and/or jQuery selectors _into_ the control, etc.

swilliamset:
Should we add access attributes directly from markup to this? This isn't the first time we've ran into attributes not being synced. It's a common problem with jQuery objects.

PR #1077, opened 2015-02-11
Bugfix for repeater list extension returning wrong rowData object in rowRenderer callback.

### LINKED ISSUE #1075: Repeater: list_rowRendered issue
The helpers object that is passed into custom row renderer function has changed with #1022 .

Previously, `helpers.rowData` was an object that represented the current row.  Now it is an array that represents *all* rows.

Previously:
screenshot 2015-02-10 18 28 14

Currently:
screenshot 2015-02-10 18 28 39

futuremint:
Adds unit tests to both list callbacks to verify the shape of the rowData object now.
Fixes #1075.

### DIFF
--- a/js/repeater-list.js
+++ b/js/repeater-list.js
@@ -379,7 +379,7 @@
 			this.viewOptions.list_rowRendered({
 				container: $tbody,
 				item: $row,
-				rowData: data
+				rowData: data.items[index]
 			}, function () {});
 		}
 	}
--- a/test/repeater-list-test.js
+++ b/test/repeater-list-test.js
@@ -129,6 +129,8 @@ define(function(require){
 				if(!hasCalled.column){
 					ok(1===1, 'columnRendered callback called upon rendering column');
 					equal((helpers.container.length>0 && helpers.item.length>0), true, 'columnRendered helpers object contains appropriate attributes');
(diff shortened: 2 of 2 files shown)


js/repeater-list.js
```diff
@@ -379,7 +379,7 @@
 			this.viewOptions.list_rowRendered({
 				container: $tbody,
 				item: $row,
-				rowData: data
+				rowData: data.items[index]
```
kevinparkerson: this can be simplified to just `subset[index]`  I was already passing in the correct "items" array as the subset, I just mistakingly put the wrong argument there >.<

kevinparkerson:
Nice catch!!!

js/repeater-list.js
cormacmccarthy: what even _is_ subset? That variable name needs to change...

kevinparkerson: These docs will change soon, but: http://getfuelux.com/extensions.html#writing-extensions-repeater-helpers

It's the "subset" of information that is being iterated over. In this case, data.items. `renderRow` is called passing the `helpers.subset` value as the `subset` argument. I suppose it could be renamed "items" in the scope of the `renderRow` function...

cormacmccarthy: Is there any case where it isn't just a convenience accessor for the data.items param?

PR #1078, opened 2015-02-11
Adding verbose to travis test task

interactivellama:
What the Fail! There's been questions on the team of figuring out which unit test actually failed in TravicCI. This should fix it. _No more F's instead of dots action going on._

Result: https://travis-ci.org/ExactTarget/fuelux/builds/50368406


cormacmccarthy:
Can you describe the symptom you were seeing and the now expected output? I need to know how to test this, or what I should now be seeing (and instead of what). Thanks!

interactivellama:
The previous task was not verbose and therefore didn't show which tests failed.

PR #1079, opened 2015-02-11
Repeater List Better Var Names

kevinparkerson:
Updates to recent repeater-list refactor for var names that are easier to understand upon viewing

### DIFF
--- a/js/repeater-list.js
+++ b/js/repeater-list.js
@@ -212,7 +212,7 @@
 				return false;
 			},
 			renderItem: function(helpers){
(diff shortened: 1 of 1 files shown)


js/repeater-list.js
```diff
@@ -212,7 +212,7 @@
 				return false;
 			},
 			renderItem: function(helpers){
-				renderRow.call(this, helpers.container, helpers.data, helpers.subset, helpers.index);
+				renderRow.call(this, helpers.container, helpers.subset, helpers.index);
```
cormacmccarthy: renderRow only accepts 3 params, you are passing in 4.

PR #1084, opened 2015-02-12
first pass at implementing phantomjs for travis

cormacmccarthy:


cormacmccarthy:
All this is provided by grunt-contrib-qunit already. Don't work late at night kids :)

PR #1085, opened 2015-02-12
corrected an event name in  wizard docs stepclick.fu.wizard -> stepclicked.fu.wizard

dkilgore:


swilliamset:
confirmed https://github.com/ExactTarget/fuelux/blob/master/js/wizard.js#L257 thx for the fix

PR #1103, opened 2015-02-17
fixes #1094 makes dist compile and test prior to release. Also prompts f...

### LINKED ISSUE #1094: Update dist and qunit process
QUnit should run on the compiled dist code, which should be generated when QUnit task is kicked off.

Everyone should be checking in their dist --OR maybe not(!).

* Update Grunt tasks to make qunit task generate the dist and then run qunit
* Update QUnit to test the dist files instead of the dev files
* ~~Update documentation to tell people to always commit their dist files~~
* ~~Update documentation to tell people how to deal with the now more frequent merge conflicts that this will result in~~

Maybe also keep the current way to run unit tests around so you can test your files while you develop on them for speed. Like a qunit-dev task or something, idk...

cormacmccarthy:
...or release type to determine new version number

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -6,8 +6,8 @@ module.exports = function(grunt) {
 	// use --no-livereload to disable livereload. Helpful to 'serve' multiple projects
 	var isLivereloadEnabled = (typeof grunt.option('livereload') !== 'undefined') ? grunt.option('livereload') : true;
 
-	// release minor or patch version. Do major releases manually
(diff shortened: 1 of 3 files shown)


interactivellama:
The bump is now occurring after the `dist` build (of `releasetest`) and therefore the `dist` files do not have the correct version. It was working correctly the two earlier times we talked. We should probably just run dist again at the end after the bump.

cormacmccarthy:
Fixed. `dist` now runs twice.

Issue #1113, opened 2015-02-19
Checkbox selection via keyboard doesn't work


computapars:
Users using keyboard navigation to checkboxes cannot select the checkbox. On focus to the checkbox, the focus is triggered but nothing shows when selected/checked via keyboard action.

interactivellama:
This makes our checkbox fail accessibility. :-( Thanks @computapars! We are planning on sunsetting the current checkbox for a control focused on styling and not JS. We may not fix this. If we don't fix it, we will have something accessible in it's place.

cormacmccarthy:
In the upcoming 3.7.0 release we are removing the JS checkbox and radio buttons entirely and replacing with a CSS only solution, which will actually fix this bug.

cormacmccarthy:
New accessible checkbox and radio buttons are now in master and will go out in the next release.

Issue #1019, opened 2015-02-02
Deprecate checkbox and radio buttons


chriscorwin:
We are deprecating `checkbox-custom` and `radio-custom`, in favor of simply using native controls.

This will be both easier to maintain into the future and has the added benefit of having behaviors that users (and developers) expect.

_Style_ the native controls to _look just like the custom controls did_ prior to deprecation, and add deprecation warnings to console similar to how the tree control does now. (rumor)

![custom-radio](https://cloud.githubusercontent.com/assets/362417/6003663/4a2f8022-aacb-11e4-9224-0195f345ed31.png)

![custom-checkbox](https://cloud.githubusercontent.com/assets/362417/6003668/5a982d24-aacb-11e4-9640-9e82b227cc57.png)

The custom controls will be removed in a future release.

mbeard:
Add a style for an indeterminate state checkbox as well

futuremint:
Example of setting the attribute: `$("#some-checkbox").prop("indeterminate", true);`.  This would be a "visual only" style on the checkbox and not effect the actual value of the input.

futuremint:
We should still support highlight and toggling in some way... maybe with a smaller plugin.

cormacmccarthy:
This actually isn't a breaking change because this can live alongside the old way.

merong12:
NAVER - http://www.naver.com/
--------------------------------------------

merongmall@naver.com 님께 보내신 메일 <Re: [fuelux] Deprecate checkbox and radio buttons (#1019)> 이 다음과 같은 이유로 전송 실패했습니다.

--------------------------------------------

받는 사람이 회원님의 메일을 수신차단 하였습니다. 


--------------------------------------------

Issue #1107, opened 2015-02-18
less/fuelux-no-namespace.less causing watch task problems


cormacmccarthy:
Slow churning infinite loop initiates while using the `grunt servefast` task. As soon as you make a file change and the watch task kicks off, the face that the no-namespace file deletes itself kicks off the next watch task, rinse, wash, repeat, forever.

interactivellama:
I'm not getting that problem, but `index.html` is broken because the client side LESS compiler is not working. I'm thinking the no namespace enhancement is causing more problems than it's worth.

What about wrapping each files contents individually in `.fuelux {}` would that solve the parent selector issue? @chriscorwin 

I don't mind it not being dry and having `.fuelux {}` at the top of 12 CSS files. I'm little less worried about MCtheme, since that is internally focused.

![screen shot 2015-02-18 at 1 54 36 pm](https://cloud.githubusercontent.com/assets/1290832/6254321/b232b9bc-b775-11e4-935c-6048a28f4118.png)

interactivellama:
Update: I am having watch problems also.

interactivellama:
Ran into this issue today, too. Looks like the source mapping is incorrect.

![screen shot 2015-02-18 at 2 21 32 pm](https://cloud.githubusercontent.com/assets/1290832/6254857/a876e30e-b779-11e4-99fa-ca7a3197fe3d.png)

chriscorwin:
Didn't this get fixed?

The fix I did seems to be in master, anyway.

cormacmccarthy:
From what I can tell, your fixes are in #1115, not master: https://github.com/ExactTarget/fuelux/pull/1115/files#diff-35b4a816e0441e6a375cd925af50752cL347

chriscorwin:
Okay, well, that' in review and should be in master soon.

PR #1115, opened 2015-02-19
Gh1019   deprecate checkbox and radio buttons

chriscorwin:
Modified styles for radio and checkbox so that they will look correct when the custom controls are deprecated.

Added a new "dev" workflow to grunt, along with a `servedev` task that bypasses building the normal `dist/` stuff, added things to `.gitignore` for the files it _does_ make.

### DIFF
--- a/less/fuelux.less
+++ b/less/fuelux.less
@@ -7,7 +7,11 @@
 @import "icons.less";
 
 // Fuel UX controls
-@import "checkbox.less";
+@import "forms.less"; // forms.less contains code shared between form elements.
+@import "checkbox.less"; // Checkbox styles that will be deprecated.
+@import "checkbox-no-js.less"; // Checkbox styles that are not dependant upon custom javascript controls.
+@import "radio.less"; // Radio styles that will be deprecated.
+@import "radio-no-js.less";  // Radio styles that are not dependant upon custom javascript controls.
 @import "combobox.less";
 @import "datepicker.less";
 @import "infinite-scroll.less";
@@ -16,7 +20,6 @@
 @import "misc.less";
 @import "pillbox.less";
 @import "placard.less";
-@import "radio.less";
 @import "repeater.less";
 @import "repeater-list.less";
 @import "repeater-thumbnail.less";
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,9 @@ dev/dev.html
 
 # screenshots
 page-at-timeout-*
+
+# dev css and maps
+dist/css/fuelux-dev.css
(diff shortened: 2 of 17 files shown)


less/fuelux.less
```diff
@@ -8,6 +8,8 @@
 
 // Fuel UX controls
 @import "checkbox.less";
+@import "radio.less";
+@import "forms.less";
```
cormacmccarthy: can we rename something like `common.less`?

cormacmccarthy:
Hey, nice, it _isn't_ a breaking change!

We'll wait until the 3.7.0 release to rip out ye ole checkbox/radio then...

In the meantime, we may want to consider some deprecation warnings or something in the console output when someone uses the old versions.

cormacmccarthy:
closes #1113, fixes #1019 and #1107

cormacmccarthy:
please rebase to master and address any merge issues. Thanks!

chriscorwin:
Assigning back to @cormacmccarthy now that I've tested the merge out and it seems okay.

interactivellama:
Please add text to the original PR comment above stating this does not work in all browsers, specifically all versions of IE (8-11) (although, I hear Windows Phone might).  We should note that `-moz-appearance` and `webkit-appearance` are non-standard CSS features and may disappear at a point in the future.

> Non-standard
> This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.

From [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/-moz-appearance),

interactivellama:
Please fix:
- Firefox compatibility
- HTML validation issues
- dev workflow changes may need to go in a different PR since it's not related to checkbox/radio.

interactivellama:
Closing until we figure out a fix.

swilliamset:
Lets keep these open if they are still in progress (even if "in a completely different way"). Owner should force push to fork when complete.

chriscorwin:
Wait again on this --- bug in radio with alignment.

interactivellama:
I'll take it when it's ready again.

chriscorwin:
It's ready, sorry.

interactivellama:
Please fix:

- HTML validation issues

chriscorwin:
Validation warnings laxed to account for header in section thing, and ARIA controls issues fixed.

Also have merged in lastest master and resolved conflicts with @cmcculloh work involving auto-linking to sections in the `index.html` file.

Ready to test one more time!

PR #1124, opened 2015-02-23
Refactor QUnit with code coverage

### LINKED ISSUE #1081: Set up Blanket/Qunit to output code coverage to console.

interactivellama:
Fixes #1081 in that it will fail if the coverage decreases. Current threshold is 68 in individual controls and 81 as an average overall. If you add functionality, you will need to add unit tests now for tests to pass.

Issues:
- The tests were being run without the test server starting up. The main server serving the files was being used for testing instead.
- Coverage cannot be checked when testing `dist`.
- Coverage cannot be tested when testing /test/*.html (no server).

Needed changes:
- QUnit runs global variables test and test on dist (when needed)
- Blanket QUnit runs source file tests with code coverage. Threshold has been set to 1%, so that it can ignored and that tests would still pass.
- Remove references to `lib` folder in watch task
- Add all tests to `serve` task and make it "do everything."
- Add a `serveslow` task to do all tests on source without building dist
- Rename qunit:simple to qunit:source for accuracy
- Add length of time to aid developers in choosing a serve task
- Remove 'feulux-' filename prefix from tests, and add tests.html as a default file when serving.
- Re-order tests, so that `blanket_qunit` will fail first. Globals hardly ever fails for obvious reasons.


Gruntfile.js
```diff
@@ -546,12 +574,14 @@ module.exports = function (grunt) {
 
 		if (!grunt.option('no-tests')) {
 			grunt.task.run(['releasetest']);
-			//delete any screenshots that may have happened if it got this far. This isn't foolproof because it relies on the phantomjs server/page timeout, which can take longer than this grunt task depending on how long saucelabs takes to run...
+			// Delete any screenshots that may have happened if it got this far. This isn't foolproof
```
cormacmccarthy: what's up with splitting this onto multiple lines?

Gruntfile.js
```diff
@@ -559,52 +589,55 @@ module.exports = function (grunt) {
...
 
-	//Fastest serve command for freely slinging code (you won't get interrupted by tests, etc).
-	grunt.registerTask('servefast', 'Serve the files. pass --test to run tests.', function () {
+	//Fastest serve command for freely slinging code (no tests will run by default).
+	grunt.registerTask('servefast', 'Serve the files (no watch), --test to run tests. (~0s)', function () {
 		grunt.task.run(['connect:server']);
 
 		if (grunt.option('test')) {
-			grunt.task.run(['qunit:simple', 'watch:simple']);
+			grunt.task.run(['connect:testServer', 'qunit:source', 'watch:source']);
```
cormacmccarthy: Why did this switch to the testServer?

Gruntfile.js
```diff
@@ -559,52 +589,55 @@ module.exports = function (grunt) {
...
-	//basically just `grunt serve` but tests default to being off
-	grunt.registerTask('servedist', 'Compile and serve everything, pass --test to run tests.', function () {
+	// same as `grunt serve` but tests default to being off
+	grunt.registerTask('servedist', 'Compile and serve everything, pass --test to run tests. (~7s)', function () {
 		grunt.task.run(['dist']);
 
-		//start up the server here so we can run tests if appropriate
+		//start up the servers here so we can run tests if appropriate
 		grunt.task.run(['connect:server']);
+		grunt.task.run(['connect:testServer']);
```
cormacmccarthy: Why start two servers? Which one is supposed to be the one the files are serving from for development? Will both serve all the same files?

test/index.html
```diff
@@ -19,37 +17,43 @@
...
+				- - - - - - - - - - - - - - - - - - */
+
+				// jQuery config
+				var jQueryPath;
+				// Feature detect
 				if('querySelector' in document
 					&& 'localStorage' in window
 					&& 'addEventListener' in window ) {
-					jQueryPath = 'bower_components/jquery/dist/jquery';
+					jQueryPath = 'bower_components/jquery/dist/jquery';	// v2.x
```
cormacmccarthy: Hrm... I'm slightly concerned it will be too easy/likely for this comment to be wrong. When we upgrade to jquery 3.x no one is going to remember to edit it. Then again, by then this whole file may be gone, so... idk.

test/tests-no-moment.js
```diff
@@ -6,7 +6,7 @@ define(function(require){
 	var $ = require('jquery');
 
 	QUnit.start(); // starting qunit, or phantom js will have a problem
-
+
```
cormacmccarthy: lol

test/tests-no-moment.js
```diff
@@ -42,23 +42,6 @@ define(function(require){
 		};
 	});
 
-	require('test/checkbox-test');
-	require('test/combobox-test');
 	require('test/datepicker-test');
```
cormacmccarthy: When does this `tests-no-moment.js` file even get called now?

test/index.html
interactivellama: I'm guessing they'll change the file structure again before that happens.

dougwilson:
> If you add functionality, you will need to add unit tests now for tests to pass.

It also means if you remove substantial parts of tested code, you have to add more tests to untested code, since that's another way to drop the percentage.

Gruntfile.js
interactivellama: Yes, they will. Here's were I made [the change](https://github.com/ExactTarget/fuelux/commit/66a8a96f386b475df2f2d2b7a8886587719b46a2). If my memory serves me right, I think it had something to do with SauceLabs. Maybe the tunneling was blocking the main server, so the testServer was set up so you can serve at the same time, since saucelabs takes so long? I'll need to look into it some more to make sure.

Gruntfile.js
interactivellama: I couldn't read it. It went off the page. I don't have line wrap on, do you? I thought we typically tried to keep code to 80 characters or something.

test/tests-no-moment.js
interactivellama: Thanks must have lost in merge manual merge conflicts. Will update in the morning.

Gruntfile.js
```diff
@@ -43,16 +43,31 @@ module.exports = function (grunt) {
...
 		}),
-		testUrl: ['http://localhost:<%= connect.testServer.options.port %>/test/fuelux.html?jquery=' + '1.9.1&testdist=true'],
 
 		//Tasks configuration
+		blanket_qunit: {
+			source: {
+				options: {
+					urls: ['http://localhost:<%= connect.testServer.options.port %>/test/?coverage=true&gruntReport'],
+					threshold: 68,
+					globalThreshold: 81
```
cormacmccarthy: So, what _exactly_ happens if this fails? Is there ever a possibility of it failing during a servefast or a normal serve, and blocking development? I understand with TDD, you write your tests first, but, we don't actually do that here to my knowledge. I'm not really ok with this being anything other than an informational tool. If it throws a warning or something that says "Your PR is likely to be rejected because it appears to not have proper unit tests" that's really neat and useful imo, but, failing out of the grunt task entirely is... unsavory to me.

Gruntfile.js
interactivellama: Yeah, I'm thinking that was reason. It means you can use 8000 for working on the project while you are sending it to Saucelabs. This decision was made when were doing regular testing of IE8  via Saucelabs and did not use the `useAvailablePort: true ` setting.

We can probably sunset the testServer for this "serving parallel servers of the same repos" edge case, but we may want to tell people to check your connect server port (as in it might be 8001 instead.) Here is a screenshot of the context below:

screen shot 2015-02-24 at 11 02 55 am

test/tests-no-moment.js
interactivellama: I've added the `?no-moment=true` URL parameter to the `test` and `releasetest` tasks.

Gruntfile.js
interactivellama: I always use `--force` when developing anyway, because if I save the file mid-line, it's guaranteed to fail. I will remove `?coverage=true` from `grunt servefast` if you want to run coverage and then you should run `grunt test` when you are done. How's that?

cormacmccarthy: That's good for `servefast`. However, getting into the habit of actively ignoring errors seems at the least odd and at the most outright dangerous (when I've seen this done before, it usually ends up causing some major problems). I've heard several people say things like "ugh, our grunt setup is useless because I always just `--force` everything". If everyone is always forcing everything anyways, why not just make that the default behaviour and then allow the non-desirable behaviour to be an intentional decision? Might be good to hear others weigh in on this though. @swilliamset @futuremint

swilliamset: sorry for the delay. this should default to a warning and not require `--force`. i do not advise using `--force` for regular usage.

PR #1125, opened 2015-02-24
Gh1105   tree disclosures

### LINKED ISSUE #1105: Disclose all tree folders function needed
Need a way to tell a tree to fully expand/disclose all folders. Basically the exact inverse of `collapse`.

- [x] Create `discloseAll` function
- [x] Create `discloseVisible` function
- [x] Create `closeAll` alias to `collapse` function for API sanity/consistency
- [x] Add kill switch for _too many branches_ issue
- [x] Add unit tests
- [ ] ~~Update documentation~~

### LINKED ISSUE #1106: `collapse` does not have a unit test and may not even be working
On the `index.html` page, if you disclose the first tree item and then call `$('#myTree1').tree('collapse');` from the console, nothing happens.

- [ ] Add unit test
- [ ] Make sure `collapse` even works

cormacmccarthy:
fixes #1105 fixes #1106


cormacmccarthy:
Travis failing because of:
image

This PR doesn't even touch that, so, I'm not sure what is going on here :|

js/tree.js
```diff
@@ -325,32 +332,156 @@
...
+				if (self.$element.find(".tree-branch.tree-open:not('.hide')").length === 0) {
+					self.$element.trigger('closedAll.fu.tree', {
+						tree: self.$element,
+						reportedClosed: reportedClosed
+					});
+					self.$element.off('loaded.fu.tree', self.$element, closedReported);
+				}
+			};
+
+			//trigger callback when all folders have reported opened
```
tweettypography: Comment should say "closed" not "opened"

js/tree.js
```diff
@@ -325,32 +332,156 @@
...
+				/*
+				* give the dev a hook to listen for exceeding disclusure limit `exceededDisclosuresLimit`.
+				* We can't do `disclosures >= upperLimit` because then if they call it again it will
+				* exit immediately unless they "reset the machine". If instead we check to see if
+				* `disclosures` is a multiple of `upperLimit`, it will run exactly `upperLimit` times
+				* while preserving (as an accurate data point for the dev) the number of `disclosures`
+				* that have occurred on this element since page load
+				*/
+				self.$element.data('disclosures', self.$element.data('disclosures') + 1);
+				if (self.options.disclosuresUpperLimit >= 1 && (self.$element.data('disclosures') % self.options.disclosuresUpperLimit) === 0) {
```
tweettypography: Why is there an event thrown here and on line 428?

js/tree.js
```diff
@@ -325,32 +332,156 @@
...
+						tree: self.$element,
+						disclosures: self.$element.data('disclosures')
+					});
+				} else if (self.options.disclosuresUpperLimit >= 1 && self.$element.data('disclosures') > self.options.disclosuresUpperLimit) {
+					/*
+					* Give the dev an out at each execution following the upper limit execution.
+					* If this is not the first time `discloseAll` has been called, this will
+					* likely fire _each round_ because the tree remembers the total disclosures since
+					* page load, _even if you've called `closeAll`. In order to prevent that, you will
+					* need to manually reset `disclosures` if you call `closeAll`:
```
tweettypography: "Manually reset `disclousures`"? Really? What's the reason that closeAll doesn't take care of this?

js/tree.js
```diff
@@ -372,11 +503,36 @@
...
+		* `ignoreRedundantOpens` to `true` will make the folder instead ignore
+		* the redundant call and stay open.
+		*/
+		ignoreRedundantOpens: false,
+		/*
+		* this is not necessarily indicative of how many layers deep the tree goes,
+		* only how many times `discloseAll` should be called before a warning trigger
+		* is fired for the dev to hook into. `discloseAll` will continue running
+		* after this many recursions unles the dev manually kilss it by setting
+		* `keep-disclosing` to `false`:
```
tweettypography: Why does it keep running rather than stopping at the limit and waiting to be told to continue?

js/tree.js
```diff
@@ -372,11 +503,36 @@
...
 		cacheItems: true,
 		folderSelect: true,
-		itemSelect: true
+		itemSelect: true,
+		/*
+		* currently calling "open" again on a folder will close it. Setting
+		* `ignoreRedundantOpens` to `true` will make the folder instead ignore
+		* the redundant call and stay open.
+		*/
+		ignoreRedundantOpens: false,
```
tweettypography: I'm not sure what the exact reason for this is

js/tree.js
swilliamset: Intention here is to help devs protect themselves. The discloseAll is recursive like in that it calls itself after the folder is loaded. If there are many layers in the folder structure this could go on for a long time but the browser would not stop it since it is not true recursion. This allows the dev to hook into the event and decided what to do with the fringe case. `keep-disclosing` is a kill switch.

js/tree.js
swilliamset: see [earlier comment](https://github.com/ExactTarget/fuelux/pull/1125/files#r25301249)

js/tree.js
tweettypography: But then why not actually stop and wait for the response?

js/tree.js
tweettypography: My question here was why `(self.$element.data('disclosures') % self.options.disclosuresUpperLimit) === 0` _and_ `self.$element.data('disclosures') > self.options.disclosuresUpperLimit`

js/tree.js
cormacmccarthy: I was wrong, you don't need to manually reset `disclosures`, unless you have `cacheItems` set to `false`, I've added a listener for that scenario that will reset `disclosures` to 0.

js/tree.js
cormacmccarthy: How exactly would you do that? We could certainly bail at that point and have them listen for that bail and just restart it if they want, but, how do we decide when to bail for them? Their disclosures may be much heavier than ours. 1 layer of theirs may be worth 100 layers of ours. The decision was made to treat this more like a warning than an error. It is non-blocking, while still allowing the dev to implement some sort of mechanism to kill it, or prompt their users for whether their users would like to continue. I've included the three line example of what it would take to turn it from a non-blocking warning into a terminator. Reversing this would just mean I'd have to set the limit to infinite (meaning it wouldn't even check it) by default because we don't want people to perceive the function itself as broken ("ugh, they chose 6, but, mine could go 300 layers deep no problem") but rather the browser as broken ("ugh, when I get to layer 30 the whole browser becomes super unresponsive, how can I kill this thing at 29?"). So, either it starts throwing the event at layer 6 by default, or it never throws the event by default. In that case it's basically 6 of one half a dozen of the other...

js/tree.js
cormacmccarthy: So that this isn't a breaking change. Calling "open" on something, should do that. However, the current API instead treats "open" as a "toggle" and will close a folder that is open if you call `openFolder` on it. This switch allows you to fix the API until 3.7.x when we can deprecate the switch and make `openFolder` behave correctly by default. One of the first problems I encountered when trying to build the `discloseAll` function was that suddenly folders were closing themselves when I was accidentally calling `openFolder` on already opened folders. This was annoying, confusing, and hard to debug.

js/tree.js
cormacmccarthy: Because I added the later today without fully thinking through the implications ;)

My original thought process doesn't really matter, sufficed to say, that I've simplified the whole thing by trashing the `%` one.

js/tree.js
tweettypography: Makes sense. Can you just add a bit more detail to the explanation? Maybe even point to a release or a github issue where this will change?

js/tree.js
tweettypography: My thought was that we would have an infinite limit by default. I guess your way _works_, just not the way I'd personally expect, so from that perspective I'll leave it up to you

cormacmccarthy: No, I actually think you might be right here. I think it is probably more intuitive to make the limit actually be a hard limit but set it to infinite. Then, make the `keep-going` switch be sort of a "ignore the limit" flag. I think I agree that this is more intuitive.

PR #1134, opened 2015-02-26
Fuelux2 intelligent dropdown cyclical dependency fix #1123

### LINKED ISSUE #1123: fuelux 2 all fails to correctly report loading because of cyclical depdency
https://github.com/ExactTarget/fuelux/blob/fuelux2/src/intelligent-dropdown.js#L9

`"fuelux/all"` needs to be remove and maybe replace with a more detailed dependency

swilliamset:
3 unit tests forward and 2 back

Unfortunately qunit appears to be preventing bubbling when bs dropdown toggle returns false where as the browser handles it fine.

Integration test handled in the browser.

@kevinparkerson this will need your v2 experience

### DIFF
--- a/src/intelligent-dropdown.js
+++ b/src/intelligent-dropdown.js
@@ -6,7 +6,7 @@
  * Licensed under the MIT license.
  */
 
-define([ "jquery", "fuelux/all"], function($) {
+define([ "jquery" ], function($) {
 
 	$(function() {
 		$(document.body).on("click", "[data-toggle=dropdown][data-direction]", function( event ) {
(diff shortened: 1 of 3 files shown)


kevinparkerson:
I don't have time to look at this at the moment. @vernak2539 would you be willing / able to take a peek?

interactivellama:
@kevinparkerson The only thing that @swilliamset is asking for is if removing `fuelux/all` will break anything. Otherwise it's just unit tests, and the rest of us can review the Qunit issues.

PR #1031, opened 2015-02-04
Added data-name for selectedItem

swbullis:
Added Ability to change step based on data-name of the wizard pane instead of just step id.

PR #1141, opened 2015-02-27
Added data-name for selectedItem

### LINKED ISSUE #607: Wizard in 3.0.0 doesn't allow for non-numeric step name (eg data-step='confirm')
By allowing non-numerical step names it makes it easier to manage changes to the flow, and to better handle dynamic insertion/removal of steps to the wizard.

cormacmccarthy:
closes #1031 fixes #607


swilliamset:
i'll review

swilliamset:
New functionality is valid and unit test covers new functionality. Possible future consideration would be to return step and name on selection when applicable.

PR #1145, opened 2015-03-03
Gh1121   make tree.openfolder only open folders instead of toggling them

### LINKED ISSUE #1121: Make tree.openFolder only open folders instead of toggling them
the `openFolder` method on the `tree` is misnamed, it is actually a toggle. Issue #1105 created a `toggleFolder` method, but in order to not be a breaking change it kept `openFolder`'s behavior as being "toggle" unless you set your options object to `{ignoreRedundantOpens: true}`.

This issue is to remove that from the `openFolder` method so that `openFolder` truly just opens the folder by default.

cormacmccarthy:
fixes #1121

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,7 +12,7 @@ define(function (require) {
 	};
 
 	// programmatically injecting this is so much easier than writing the html by hand 376 times...
-	$('h1[id], h2[id], h3[id], h4[id], h5[id], h6[id], dt[id]').each(function(i){
+	$('h1[id], h2[id], h3[id], h4[id], h5[id], h6[id], dt[id]').each(function (i) {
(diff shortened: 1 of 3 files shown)


interactivellama:
Reviewed, but should wait until 3.7 release to merge due to an API breaking change.

cormacmccarthy:
:+1:

swilliamset:
@cormacmccarthy or @interactivellama please rebase and retest so able to merge

interactivellama:
@cormacmccarthy I originally cloned this down and phantomJS timed out. I thought it was my setup and so I ran the tests again and they all passed, so I merged. Do you know if there is a history of test fails with usage of `disclose`. Should we put a counter/limit on our test datasource to prevent time outs?

https://travis-ci.org/ExactTarget/fuelux/builds/58018262#L588

PR #1181, opened 2015-03-19
Ignore clicks on disabled menu items

jwarby:
The Select component shouldn't allow a menu option with the 'disabled' class to be selected (see [Disabled menu options](http://getbootstrap.com/2.3.2/components.html)).

It is easy enough to bind a click handler to Select component's DOM element to prevent this behaviour, but there is no way to achieve this for dynamically inserted Select components.  Many MVC frameworks use event delegation to handle events for dynamically inserted DOM content, so there is no way to prevent this behaviour in such frameworks (at least not without breaking the framework's conventions by hacking in a workaround).

[This jsfiddle](http://jsfiddle.net/warby_/an0ws9nq/1/) might help to explain the issue a bit more clearly.

### DIFF
--- a/src/select.js
+++ b/src/select.js
@@ -34,6 +34,11 @@ define(function(require) {
 
         itemclicked: function (e) {
             this.$selectedItem = $(e.target).parent();
+
+            if (this.$selectedItem.is('.disabled')) {
+                return;
+            }
+
             this.$hiddenField.val(this.$selectedItem.attr('data-value'));
             this.$label.text(this.$selectedItem.text());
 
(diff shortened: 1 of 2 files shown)


interactivellama:
Thanks, @jwarby. Will review.

jwarby:
I just noticed this is targeted to the 3.6.5 milestone, but my pull request is for FuelUX 2 (as is my other outstanding pull request) - just thought I should mention...

interactivellama:
Sorry, thanks for the clarification.

Issue #1166, opened 2015-03-06
repeater dataSource to support array


swilliamset:

kevinparkerson:
It doesn't make sense to support just an array instead of a function, since the function / callback pattern is a common convention across several fuelux controls with a `dataSource`. Instead, I've made the repeater function even without a dataSource and be a lot more lenient with what's passed back in the callback. @swilliamset thoughts on this decision?

swilliamset:
```If( dataSource.constructor === Array ) {
  // determine length page etc
  // call callback
}```

cormacmccarthy:
We will be addressing this in an upcoming release. Thank you!

futuremint:
This would be a great idea, however we're not going to add this unless its an explicit ask from a scrum team.

PR #1182, opened 2015-03-20
Repeater better defaults

### LINKED ISSUE #1167: repeater to render with no dataSource
currently errors and/or infinitely loads

kevinparkerson:
Fixes #1167 by allowing repeater to load if no dataSource has been provided. Also attempts to address issue #1166 by making the repeater still function if bad data or no data is passed back in the dataSource, and provides better defaults for itemization and pagination in the event of poor data scenarios.

Also fixes a random bug I noticed with the `deselected` event in repeater-list

### DIFF
--- a/js/repeater-list.js
+++ b/js/repeater-list.js
@@ -345,7 +345,7 @@
 				if ($item.hasClass('selected')) {
 					$item.removeClass('selected');
 					$item.find('.repeater-list-check').remove();
-					$item.$element.trigger('deselected.fu.repeaterList', $item);
+					self.$element.trigger('deselected.fu.repeaterList', $item);
 				} else {
 					if (self.viewOptions.list_selectable !== 'multi') {
 						self.$canvas.find('.repeater-list-check').remove();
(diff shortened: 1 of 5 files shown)


mbeard:
@kevinparkerson - does this allow for rendering of the column headers prior to data being loaded?

kevinparkerson:
@mbeard not in this PR, but that is something we could do for a new issue / PR. I think it would be the addition of a "renderColumnHeaders" function to repeater-list

interactivellama:
This is working great with repeater-list. When I use repeater thumbnail view as default, I do not get `no items found` on the repo's index.html page, even though it's specified in the initiation options.

screen shot 2015-03-23 at 4 50 37 pm

kevinparkerson:
There was some bad logic from the plugin API translation in the repeater-thumbnail `before` function that would not render the thumbnail container in certain circumstances, so the `noItemsHTML` never appeared. Good catch!

interactivellama:
Font size needs to be over written if empty. It's currently .1px, so "no items found" is present, but not readable.

screen shot 2015-03-24 at 9 17 11 am

PR #1184, opened 2015-03-23
Datepicker type colors and border radius inconsistent with rest of fuel ux

### LINKED ISSUE #1088: Datepicker: type colors and border radius inconsistent with rest of Fuel UX
Type colours are mysteriously being overridden, rather than inheriting bootstrap defaults.

Border radius on the buttons is also being overridden to be 0 instead of the default 4px.

chriscorwin:
Fixes #1088

### DIFF
--- a/less/datepicker.less
+++ b/less/datepicker.less
@@ -64,8 +64,6 @@
 				&.current-day {
 					span {
 						border: 1px solid #333;
-						b, button {
-						}
 					}
 				}
 
--- a/less/datepicker.less
+++ b/less/datepicker.less
@@ -38,7 +38,7 @@
 
 			td {
 				b, button {
-					color: #231f20;
+					color: #333;
 					display: inline-block;
 					font-weight: normal;
 					height: 30px;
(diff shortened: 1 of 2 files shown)


chriscorwin:
I should mention that I did some JS work in this to add `.first` and `.last` classnames on the `.next-month` and `.last-month` table cell elements, in order to have them appear as one "bar" with rounded corners on each end.

screenshot 2015-03-23 10 25 39

less/datepicker.less
```diff
@@ -60,18 +62,30 @@
 				}
 
 				&.current-day {
-					b, button {
+					span {
 						border: 1px solid #333;
+						b, button {
```
kevinparkerson: Why has this border styling been transferred to the `span` vs the `b, button`? It appears to look ok like that, just was curious about the reasoning. Might as well remove this `b, button` definition then since it is empty.

chriscorwin:
The button is subtly different size than the span, so when the outline for selected is on it, the background colour of the span sticks out further than the button's border.

Here's how it was:

screenshot 2015-03-23 12 36 18

kevinparkerson:
Ah yes... there's a reason the button is smaller (strikethrough or some other purpose, forgot) but I agree having around the full span makes more sense for the "current day" outline.

PR #1185, opened 2015-03-23
Checkbox refactor

mbeard:
Refactoring of checkbox

### DIFF
--- a/index.html
+++ b/index.html
@@ -54,10 +54,12 @@
 				<section id="checkboxes-block">
 					<div class="checkbox">
 						<label class="checkbox-custom" data-initialize="checkbox" id="myCustomCheckbox1">
-							<input class="sr-only" type="checkbox" value="">
+							<input class="sr-only" type="checkbox" value="" data-toggle="#myCheckboxToggleContainer">
 							<span class="checkbox-label">Custom checkbox unchecked on page load</span>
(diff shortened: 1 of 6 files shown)


mbeard:
I disagree.  We shouldn't support setting the checked appearance without setting the checked state.  Why would we ever want those to be out of sync.  Same for disabled.

swilliamset:
My statement was assuming the function setting the class would insure the control state was synced but i see how the existing function name could cause that confusion.

swilliamset:
@sjames include undocumented methods no longer available in the release notes

interactivellama:
The following *undocumented* public methods have been removed from checkbox:
setState, _toggleCheckedState, _toggleDisabledState, _setDisabledClass, and _setCheckedClass 

Additionally, the radio and checkbox controls are now bound to the parent element with the respective `.radio` and `.checkbox` classes and no longer to the `<input>` element.

@mbeard Anything else that you would like to see in the release notes?

chriscorwin:
Highlight isn't working.

The rule to add the background needs changed.

Right here:

https://github.com/mbeard/fuelux/blob/master/less/checkbox.less#L16


...change it to this:

`&.checked label.checkbox-custom, label.checked.checkbox-custom {`

PR #1189, opened 2015-03-24
Replace `hide` class with `hidden`

### LINKED ISSUE #1002: hide should be changed to hidden
From BS documentation:

> .hide is available, but it does not always affect screen readers and is deprecated as of v3.0.1. Use .hidden or .sr-only instead.

Things that have "hide":

1. Scheduler (7 occurrences in getfuelux.com) `<div class="form-group repeat-panel repeat-weekly repeat-days-of-the-week hide" aria-hidden="true">`
1. Tree (4 occurrences in getfuelux.com) `<li class="tree-branch hide" data-template="treebranch" role="treeitem" aria-expanded="false">`

All of fuelux should be evaluated for this issue.

interactivellama:
Across markup, tests, and JS. Fixes #1002.

### DIFF
--- a/index.html
+++ b/index.html
@@ -107,9 +107,9 @@
 							<span class="checkbox-label">Custom checkbox toggles by id</span>
 						</label>
 					</div>
-					<div aria-hidden="true" class="checkboxToggle alert alert-warning hide">This message's visibility toggles by class with a checkbox above.</div>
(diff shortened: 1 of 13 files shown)


swilliamset:
@cmcculloh as the creator of #1002 could you review this?

PR #1205, opened 2015-03-31
Repeater Highlight Sorted Columns Default Value

### LINKED ISSUE #1201: Repeater List list_highlightSortedColumn option needs to be defaulted to true
Might need to flip that to true, since in most cases highlighting the sorted column is desired.

NOTE: talk to design about this as well

kevinparkerson:
Fixes #1201: Changing default list_highlightSortedColumns value to true


interactivellama:
Thanks, Kevin!

PR #1207, opened 2015-04-01
Correct style for checkbox/radio without a text label

### LINKED ISSUE #1196: Refactor misc.less to be less greedy in it's cascade
This is a screen shot from a [feature branch test page](http://mctheme-styleguide-sync.herokuapp.com/examples/fuelux/fuelux.html#Input-groups). This is the first time I put all the markup for MCtheme within `.fuelux` wrapper.

screenshot 2015-03-25 21 49 16

The problem is caused by:
```
.input-group-addon {
	padding: 6px 18px;
}
```
which becomes
```
.fuelux .input-group-addon {
	padding: 6px 18px;
}
```
in `less/misc.less`.

@cormacmccarthy Do you know why this line was added back in Oct? I'm just looking for background.

The other classes such as `input, textarea` should probably be wrapped in the name of the control (ex- `.placard .input`), since I think it's common practice to wrap the entire page in `.fuelux`.

interactivellama:
- Fixes #1196.
- Removes misc.less and places existing LESS inside control classes to avoid CSS rule collision and play nice with everyone since `.fuelux` is often applied to the `html` tag.
- Added checkbox/radio (no-JS) example within `.input-group-addon` markup, so we can monitor what that looks like going forward.
- Made example for inline radio and checkboxes use `span`, since it may be within a `span` container and a `div` within a `span` is a validation no-no.

I'm going to call this a style breaking change, since if you were using `.glass` outside of `placard` then this is going to break that. However, that class was never documented and therefore this isn't a breaking change, and I'm assigning it to a bugfix release.


swilliamset:
Looks like a refactor of the `misc.less` file hitchhiked into this request. Please update the branch minus those changes.

PR #1216, opened 2015-04-06
Scheduler Recurrence Fix

### LINKED ISSUE #1080: Scheduler Recurrence Pattern Issue
The scheduler recurrence pattern is never reset when you use `setValue`, which means if you call `setValue` twice with different recurrence patterns, it will only override what has been set previously. Example:

```js
var schedule1 = {
	startDateTime: '2014-03-31T03:23+02:00',
	timeZone: {
		offset: '+02:00'
	},
	recurrencePattern: 'FREQ=WEEKLY;BYDAY=WE;INTERVAL=1;COUNT=3;'
};

var schedule2 = {
	startDateTime: '2014-03-31T03:23+02:00',
	timeZone: {
		offset: '+02:00'
	},
	recurrencePattern: 'FREQ=WEEKLY;BYDAY=WE;INTERVAL=1;'
};

$('#myScheduler').scheduler('value', schedule1);
$('#myScheduler').scheduler('value', schedule2);
console.log($('#myScheduler').scheduler('value'));

//recurrence will be "FREQ=WEEKLY;BYDAY=WE;INTERVAL=1;COUNT=3;"
//instead of "FREQ=WEEKLY;BYDAY=WE;INTERVAL=1;" as expected
```

kevinparkerson:
fixes #1080 by adding else condition when COUNT and UNTIL aren't present

### DIFF
--- a/js/scheduler.js
+++ b/js/scheduler.js
@@ -418,7 +418,6 @@
 				startDate = temp[0];
 
 				if (temp[1]) {
-					startTime = temp[1];
 					temp[1] = temp[1].split(':');
 					hours = parseInt(temp[1][0], 10);
 					minutes = (temp[1][1]) ? parseInt(temp[1][1].split('+')[0].split('-')[0].split('Z')[0], 10) : 0;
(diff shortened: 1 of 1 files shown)


interactivellama:
Works as expected. Thanks, Kevin!

PR #1217, opened 2015-04-06
Unit test fix

### LINKED ISSUE #1215: Unit tests don't work in browser
The current unit test configuration does not work in browser due to some bad paths.

kevinparkerson:
fixes #1215 by modifying require paths to './' from '../' since baseUrl has already been set to '../'


interactivellama:
Thanks Kevin for finding this. We were all using `localhost`'s root for `fuelux/`.

PR #1223, opened 2015-04-07
Issue 1194 repeater frozen column

BenjaminNeilDavis:
Creation of new frozen plugin for repeater

### DIFF
--- a/js/repeater-list.js
+++ b/js/repeater-list.js
@@ -139,6 +139,57 @@
 			});
 		};
 
+		$.fn.repeater.Constructor.prototype.list_setFrozenColumns = function () {
+			var frozenTable = this.$canvas.find('.table-frozen');
+			var $table = this.$element.find('.repeater-list table');
+			var repeaterWrapper = this.$element.find('.repeater-list');
+			var numFrozenColumns = this.viewOptions.list_frozenColumns;
+
+			if (frozenTable.length < 1) {
+				//setup frozen column markup
+				//main wrapper and remove unneeded columns
+				var $frozenColumnWrapper = $('<div class="frozen-column-wrapper"></div>').insertBefore($table);
+				var $frozenColumn = $table.clone().addClass('table-frozen');
+				$frozenColumn.find('th:not(:lt('+ numFrozenColumns +'))').remove();
+				$frozenColumn.find('td:not(:nth-child(n+0):nth-child(-n+'+ numFrozenColumns +'))').remove();
+
+				//need to set absolute heading for vertical scrolling
+				var $frozenThead = $frozenColumn.clone().removeClass('table-frozen');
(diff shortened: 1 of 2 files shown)


js/repeater-frozen.js
```diff
@@ -0,0 +1,534 @@
...
+			frozen_noItemsHTML: 'no items found',
+			frozen_selectable: false,
+			frozen_sortClearing: false,
+			frozen_rowRendered: null,
+			frozen_frozenColumns: 0
+		});
+
+		//EXTENSION DEFINITION
+		$.fn.repeater.viewTypes.frozen = {
+			cleared: function () {
```
kevinparkerson: This should be removed altogether unless something will reside within the function

js/repeater-frozen.js
```diff
@@ -0,0 +1,534 @@
...
+					options.sortProperty = this.frozen_sortProperty;
+				}
+				return options;
+			},
+			initialize: function (helpers, callback) {
+				this.frozen_sortDirection = null;
+				this.frozen_sortProperty = null;
+				callback();
+			},
+			resize: function () {
```
kevinparkerson: This should be removed altogether unless something will reside within the function

kevinparkerson:
Looking pretty good! I noticed a few minor issues.

1) In the dev.html page, the repeater's footer content was extending beyond the footer itself. This appeared to resolve itself when resizing the browser, so it's probably a fluke of the repeater's sizing method, which can be solved by calling "resize" at the appropriate time. Doubt it's due to the plugin, but wanted to make note of it.

2) In Firefox, the frozen column header is missing a right border. There is also a little wonkiness with the frozen column shifting around a bit.

js/repeater-list.js
```diff
@@ -202,6 +254,11 @@
 							self.list_positionHeadings();
 						}
 					});
+					if (self.viewOptions.list_frozenColumns) {
+						helpers.container.on('scroll.fu.repeaterFrozen', function () {
```
kevinparkerson: This event should be `scroll.fu.repeaterList`. Probably a symptom of moving code from the repeater-frozen file into the repeater-list plugin

kevinparkerson:
Commented on the event naming, outside that looks good!

kevinparkerson:
Update: we might have to consider row selection as well here

BenjaminNeilDavis:
To use the frozen columns:
- Use the list view option `list_frozenColumns: n ` where n is the number of columns to freeze. These must be the columns on the farthest left.
- The frozen columns table is set to table-layout: fixed; which helps with the sizing of the frozen column and to keep from changing the sizes of the columns between page switches. With this enabled you can set the width of the columns with something like this:
```
table thead tr th, .repeater-list-heading {
width: 200px;
}
```

PR #1233, opened 2015-04-11
Remove Creator section from README

adamalex:
I may have created the repository but you have all long surpassed the contributions I've made, and Fuel UX is a much bigger deal than I could have ever imagined! I appreciate the team leaving this credit in place for so long. This seems like as good a time as any to retire this section from the README.


interactivellama:
Would you be interested in being moved to the [Contributors section](https://github.com/ExactTarget/fuelux/blob/master/DETAILS.md#contributors). That would remove you off the main README.

adamalex:
Looks like I'm already there, and that's just fine with me.

interactivellama:
Done.

PR #1235, opened 2015-04-13
Repeater disabling methods

### LINKED ISSUE #1212: Repeater controls not disabling during data load
When you change pagination view on repeater and the ajax call is executing, you are able to still paginate and sort via dropdown and search.

When you interact with a filter, all filters should disable until interaction resolves.

Here is the original issue in MCAS that is generating this ORCA request:

> qa keeps coming back with issues relating to repeater controls being clickable while dataSources are loading and seeing repeated data. Is this something that should be the app's responsibility to handle?  I've already implemented this in a few spots, but there are a lot of edge cases with all of our filtering/sorting/etc. Could the fuel controls just disable all buttons on its control while the datasource is loading or ensure that the container is empty before appending the datasource response?

kevinparkerson:
Fixes #1212 by adding a disable and enable method, which disables/enables default repeater header and footer controls when called. Additionally, these methods are called prior to rendering and after render completion. Events are thrown within these methods for plugin and app timing purposes.

### DIFF
--- a/js/repeater.js
+++ b/js/repeater.js
@@ -231,7 +231,7 @@
 		enable: function() {
 			var disabled = 'disabled';
 			var enable = 'enable';
-			var noMore = 'no-more';
+			var pageEnd = 'page-end';
 
 			this.$search.search(enable);
 			this.$filters.selectlist(enable);
@@ -240,10 +240,10 @@
 			this.$primaryPaging.find('.combobox').combobox(enable);
 			this.$secondaryPaging.removeAttr(disabled);
 
-			if(!this.$prevBtn.hasClass(noMore)){
+			if(!this.$prevBtn.hasClass(pageEnd)){
 				this.$prevBtn.removeAttr(disabled);
 			}
-			if(!this.$nextBtn.hasClass(noMore)){
+			if(!this.$nextBtn.hasClass(pageEnd)){
 				this.$nextBtn.removeAttr(disabled);
 			}
 
@@ -439,8 +439,8 @@
 		pagination: function (data) {
 			var act = 'active';
 			var dsbl = 'disabled';
-			var noMore = 'no-more';
 			var page = data.page;
+			var pageEnd = 'page-end';
 			var pages = data.pages;
 			var dropMenu, i, l;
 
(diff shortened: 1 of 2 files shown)


interactivellama:
I'd like to see something a little descriptive than ".no-more".

Maybe, `pagination-complete` or something?

https://github.com/ExactTarget/fuelux/pull/1235/files#diff-e6bc01b17b5e038ff8e1ace1c0c0fbe6R442

kevinparkerson:
@interactivellama I changed it from ".no-more" to ".page-end," which is the best I could come up with :|

interactivellama:
@kevinparkerson I'm sorry, I forget to mention this yesterday. Please add the disable/enabled buttons to index.htm/index.js

js/repeater.js
```diff
@@ -211,6 +211,46 @@
...
+			var disabled = 'disabled';
+
+			this.$search.search(disable);
+			this.$filters.selectlist(disable);
+			this.$views.find('label').attr(disabled, disabled);
+			this.$pageSize.selectlist(disable);
+			this.$primaryPaging.find('.combobox').combobox(disable);
+			this.$secondaryPaging.attr(disabled, disabled);
+			this.$prevBtn.attr(disabled, disabled);
+			this.$nextBtn.attr(disabled, disabled);
```
cormacmccarthy: pretty sure these should be .prop(disabled, true) instead of attr(disabled, disabled)

http://stackoverflow.com/questions/5874652/prop-vs-attr

js/repeater.js
```diff
@@ -211,6 +211,46 @@
...
+
+			this.$search.search(enable);
+			this.$filters.selectlist(enable);
+			this.$views.find('label').removeAttr(disabled);
+			this.$pageSize.selectlist('enable');
+			this.$primaryPaging.find('.combobox').combobox(enable);
+			this.$secondaryPaging.removeAttr(disabled);
+
+			if(!this.$prevBtn.hasClass(pageEnd)){
+				this.$prevBtn.removeAttr(disabled);
```
cormacmccarthy: this.$prevBtn.prop(disabled, false);

Do _not_ use removeProp() (see https://api.jquery.com/removeProp/#entry-longdesc)

js/repeater.js
```diff
@@ -211,6 +211,46 @@
...
+			}
+			if(!this.$nextBtn.hasClass(pageEnd)){
+				this.$nextBtn.removeAttr(disabled);
```
cormacmccarthy: same as 244

PR #1245, opened 2015-04-17
add tree-folder-select class by default

### LINKED ISSUE #1243: add tree-folder-select programmatically by default
The default value for `folderSelect` is true. However the class `tree-folder-select` needs to be added to the tree container in order for the CSS to be correct.

We should add the `tree-folder-select` class if it is not present and `folderSelect` is set to true.

interactivellama:
Fixes #1243

### DIFF
--- a/index.html
+++ b/index.html
@@ -2306,27 +2306,8 @@
 		<section id="tree">
 			<h2>Tree</h2>
 			<div class="thin-box">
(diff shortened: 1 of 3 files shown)


swilliamset:
@interactivellama Please reset PR to separate fix from additional index refactoring

interactivellama:
Updated with your request for scoped commits before you would review it.

PR #1250, opened 2015-04-20
Add AMD bundle to dist build

interactivellama:
This adds a mangled JS file to the `dist/js` folder which contains Fuel UX Bundled AMD (Require.JS, jQuery, Bootstrap, Moment.js and Fuel UX). 

Approx. size when gzipped: 124KB.

Testing task (qunit:validate-distBundledAMD) has also been added to double check that concatenating and mangling does not break anything. Same actual control tests are being used. This test runs on release.

It also includes a module/define name called `fuelux-bundled-amd` for convenience for those developers that don't like a lot of typing.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -271,6 +271,11 @@ module.exports = function (grunt) {
 				options: {
 					urls: ['http://localhost:<%= connect.testServer.options.port %>/test/?testdist=true']
 				}
+			},
+			distBundledAMD: {
+				options: {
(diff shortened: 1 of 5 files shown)


kevinparkerson:
I'm ok with this technology stack bundle, but it should be in a different repo, not in Fuel UX. Otherwise it pollutes the Fuel UX library and confuses what it actually is. We should keep dist simple as possible for those using this particular library outside of our company.

kevinparkerson:
We're going to make a separate repo for this. I will do the work by next Monday.

Issue #1237, opened 2015-04-14
Ability to have x, N.x, and N.N.x releases on CDN


cormacmccarthy:
Use cases:

* Jsfiddles, codepens, etc
* Documentation sites (getfuelux.com, pages, etc)
* Small projects (non-enterprise)

Lowers cost of maintaining our disparate documentation efforts. Lowers cost of maintenance for small-potatoes sites where the cost of updating after each release is greater than the cost of fixing something when you happen to notice a release broke it.

cormacmccarthy:
Talking with "the Aarons" right now to get our CDN configured correctly to allow for this. Will require uploading to a few different places on the CDN with each release instead of just one. Will require some CDN cache flushes at release time. Will require custom expire headers on *.x CDN targets to be much lower than the year we have them set for right now (one day? one week?).

tomByrer:
[jsDelivr CDN](http://www.jsdelivr.com/) already has [versioning including `latest`](https://github.com/jsdelivr/jsdelivr#version-aliasing) built-in, with loads more features; why not use that instead?

interactivellama:
Thanks for putting that on our radar, Tom. It's working for me: http://cdn.jsdelivr.net/fuelux/latest/js/fuelux.min.js

This works wonderfully, too. http://cdn.jsdelivr.net/g/jquery,bootstrap,fuelux

PR #1251, opened 2015-04-21
Use jsDelivr as to load webpage assets

tomByrer:
2 fewer DNS lookups, auto-updating minor versions, & save a few TCP packets saved thanks to better gzip compressing concocted files.
ref: https://github.com/ExactTarget/fuelux/issues/1237#issuecomment-94459432

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,15 +29,11 @@ Read the [Getting started page](http://getfuelux.com/getting-started.html) for m
 
 Add `fuelux` class to the portion of the page using Fuel UX as seen [here](https://github.com/exacttarget/fuelux/blob/master/DETAILS.md#using-fuel-ux).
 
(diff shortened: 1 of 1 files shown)


interactivellama:
I like this. I will review it with our team today. 

Also this [grunt task](https://github.com/ExactTarget/fuelux/blob/master/Gruntfile.js#L373) will need to be modified (or possibly add another regex find and replace task), since we automate changing the version number at release, so we don't have to touch the readme.md files manually.

interactivellama:
I'm sorry, Tom, but we are too much aligned with our current CDN provider and use it in production to depend on another CDN that we are not that familiar with. We plan on using your URL in JSfiddle and some documentation from time to time. This looks like a really cool project, and I hope it grows!

tomByrer:
Thanks.

Issue #1253, opened 2015-04-21
Scrollbar in repeater bleeds into header 


swilliamset:
Because of the viewport positioning and sizing the scrollbar is not cleanly/natively positioned with the records and overlaps with the header content.

![unnamed](https://cloud.githubusercontent.com/assets/347227/7257193/cae1237c-e821-11e4-9d52-35907daa8d12.png)

BenjaminNeilDavis:
This will be fixed with my pull request for issue #1254

interactivellama:
This is a different issue. This design bug still exists after the fix to `.scrolling`

interactivellama:
I just remembered some more history on this. Before you joined the team, Stephen. Kevin and I asked design about this issue in the large pattern library meeting. It would have been May or June of 2014.

I'm still looking into this, but we may be spinning our wheels and did have design approval previously. If things have changed and design has said this is important, than we can try to accommodate for that. 

Looking into creating another view / options override.

jamin-hall:
Ok. Good to know. I was having a hard time remembering those conversations, but it’s probably because I wasn’t there. :) I think this is something we should definitely fix. It’s a small detail, but it feels off. As to how high we prioritize and when we fix it, I think there’s room for discussion.

This example feels off, but I think we can live with temporarily while we find a fix.
[cid:BFA06D17-DB52-425E-8115-D6BF2EEF1054]



This example feels broken and we should fix as soon as possible. Might seem superficial, but if a user sees this it communicates a lack of polish and plants a seed of mistrust.
[cid:B2038122-2417-406D-89EA-D5FDBBAF9954]

interactivellama:
Thanks, Jamin. I'm going to remove the high priority tag based on your feedback, but keep it in a milestone, so that it doesn't get lost.

PR #1255, opened 2015-04-22
Allow repeater to scroll when needed

### LINKED ISSUE #1254: Repeater-list not scrolling like it should
After the addition of frozen columns a class was moved causing it to not scroll properly.

BenjaminNeilDavis:
Everything was wrapped in .scrolling on the canvas. When the repeater actually needed to scroll it was being overwritten because .scrolling was not being called correctly. Issue that came from my changes for frozen columns.

Fixes #1254 and #1253


interactivellama:
Thanks for the fix, Ben!

PR #1263, opened 2015-04-24
Add radio markup deprecation notice, update markup

### LINKED ISSUE #1261: radio check method broken if documented markup is used
Out test/dev markup does not match our documentation, so our radio methods are failing if the documented markup is used. I believe this only affects 3.7.0-3.7.1. Previously releases should not be affected.

This works:

```
      <div class="radio">
        <label class="radio-custom" data-initialize="radio" id="myCustomRadio1">
          <input class="sr-only" name="radio1" type="radio" value="" id="radio1">
          <span class="radio-label">Custom radio unchecked on page load</span>
        </label>
      </div>
```
https://github.com/ExactTarget/fuelux/blob/master/index.html#L725

This does NOT work:
```
<div class="radio" data-initialize="radio" id="myCustomRadio">
  <label class="radio-custom">
    <input class="sr-only" name="radioEx1" type="radio" value="option1">
    Browser-independent radio unchecked on page load
  </label>
</div>
```
http://getfuelux.com/javascript.html#radio

~~The plan is to release an update to the code that allows the documented markup to work, since that was the intent of @mbeard 's refactor.~~

**UPDATE:** `data-initialize="radio" id="myCustomRadio"` should *always* be on the label, so that we are consistent with inline and block. Then, the only difference is the `.radio` CSS class placement.

interactivellama:
Fixes #1261.

### DIFF
--- a/_includes/js/radio.html
+++ b/_includes/js/radio.html
@@ -14,11 +14,16 @@
   <h2 id="radio-usage">Usage</h2>
 
   <h3 id="radio-usage-javascript">Via JavaScript</h3>
-  <p>Call the radio control via JavaScript:</p>
-  {% highlight js %}$('#myRadio').radio();{% endhighlight %}
+  <p>Call the radio control via JavaScript on the <code>label</code>. The <code>div.radio</code> container is only present for block level stying:</p>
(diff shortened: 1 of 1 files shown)


kevinparkerson:
Looks good

PR #1269, opened 2015-05-02
Organizing documents tree

MJ-Vakili:
Hi,
Description "deselected.fu.tree event" is incorrect.
"opened.fu.tree" not exist in code."disclosedFolder.fu.tree" is true.
add description for "updated.fu.tree" event.


interactivellama:
Thanks you for the corrections. They are very much appreciated.

PR #1282, opened 2015-05-07
Revert breaking checkbox markup change for 3.7.2 release

interactivellama:
Due to an internal timeline, the team has decided to revert the breaking change in markup for checkbox, so that we can release 3.7.2 today instead of 3.8.0.

For more background information on the markup changes that were committed into master and now reverted, please review the [3.7.0 release notes](https://github.com/ExactTarget/fuelux/releases/tag/3.7.0).


swilliamset:
Worked with @interactivellama to make the revert.

Essentially looked like this
```
git checkout master
git revert -m 1 f475eab
```

PR #1286, opened 2015-05-11
call pause (clears timeout) first when destroying

swilliamset:


futuremint:
Cool, I'll pull this down and profile it to make sure it fixes the problem.

futuremint:
This now properly idles the interpreter after calling 'destroy' on all loaders.

PR #1289, opened 2015-05-11
Base64 encode Fuel UX WOFF font into CSS file

interactivellama:
Convert WOFF Fuel UX font to base64 and store in CSS file.

For more information on fonts visit: https://css-tricks.com/snippets/css/using-font-face/

@futuremint We can do the same with BS3's glyphicon font, but that will probably be a rucksack issue @kevinparkerson.


swilliamset:
Targeted the wrong branch.

PR #1291, opened 2015-05-11
Base64 encode WOFF FuelUX font to reduce requests

interactivellama:
Convert WOFF Fuel UX font to base64 and store in CSS file.

For more information on fonts visit: https://css-tricks.com/snippets/css/using-font-face/

@futuremint We can do the same with BS3's glyphicon font, but that will probably be a rucksack issue @kevinparkerson.


interactivellama:
Follow up: From what I can ascertain from Chrome Timeline. The base64 decode may be adding 10-20ms to rendering time (It's within multiple reload margin of errors though).

Right now, network download from Akaimai of [WOFF file](http://www.fuelcdn.com/fuelux/3.7.2/fonts/fuelux.woff) ranges from 225ms to 900ms (although I am on a plane right now with pretty bad latency). 

[HTTPS requests](http://www.fuelcdn.com/fuelux/3.7.2/fonts/fuelux.woff) from Akaimai are taking 500ms to 1.5s (but then again, I'm on a plane).

We probably shouldn't be encoding large fonts, but this small one seems like a good idea for optimization.

Issue #1270, opened 2015-05-02
Pre-checked / selected radios don't toggle elements


SuperSeb92:
Considering the following:

    <label data-initialize="radio">
        <input class="checked" data-toggle=".radioToggle" checked="checked" type="radio" value="option1">
        Toggle class element
    </label>
    
    <div class="radioToggle">Toggled</div>

I'd expect to see the toggled div, without having to un-check and re-check, is this intended behaviour?

I'm currently triggering this manually using $('.fuelux input[data-toggle]:checked').trigger('change');

interactivellama:
This may be linked to #1287. Just wanted to connect the two.

Looks like we are not executing any class changes on init, only in `resetGroup` and `setCheckedState`

Issue #1287, opened 2015-05-11
Checked radio on init only works if it is last


interactivellama:
If you rearrange the checked and unchecked radios on the `index.html`. `checked` doesn't work at init if there are additional unchecked radios.

```
<div class="radio">
          <label class="radio-custom" data-initialize="radio" id="myCustomRadio2">
            <input checked="checked" class="checked sr-only" name="radio1" type="radio" value="" id="radio2">
            <span class="radio-label">Custom radio checked on page load</span>
          </label>
        </div>
        <div class="radio">
          <label class="radio-custom" data-initialize="radio" id="myCustomRadio1">
            <input class="sr-only" name="radio1" type="radio" value="" id="radio1">
            <span class="radio-label">Custom radio unchecked on page load</span>
          </label>
        </div>
```

![screenshot 2015-05-11 13 58 12](https://cloud.githubusercontent.com/assets/1290832/7571493/ed2ee18e-f7e5-11e4-8263-85658fbd7805.png)

PR #1309, opened 2015-05-21
Allow radio button by default to be checked if it is not last

### LINKED ISSUE #1306: Scheduler Monthly/Yearly Radio Buttons will not default checked correctly
Scheduler Monthly and Yearly Radio Buttons choices won't default select properly unless the selected radio button is the last one in the group in the HTML. There is a bug with the calling of the ResetGroup() function for each radio button created in the group upon SetInitialState(). This means every call to create a radio button unselects any previous ones in the group. If you set any radio button in a group except the last on to default selected it won't appear selected even thou the hidden radio button is indeed selected. Even your demo page displays the problems notice how nothing is selected in the reoccurring demo when you switch to Yearly reoccurring but the hidden radio actually is selected and you can't see it? This is going to be my fourth bug report with the exact code to change on the scheduler any kind of timeframe?  For this one you need to remove the ResetGroup() function call. It appears to serve no purpose other than to break the initialization because SetChecked() executes that exact same code that's in ResetGroup.....again

kevinparkerson:
Fixes #1306 by correcting bug in radio control where `resetGroup` was called every time `setCheckedState` was called, regardless if whether the current radio was being checked or unchecked

### DIFF
--- a/js/radio.js
+++ b/js/radio.js
@@ -84,15 +84,15 @@
 		},
 
 		setCheckedState: function(element, checked) {
-			// reset all items in group
-			this.resetGroup();
-
 			var $radio = element;
 			var $lbl = $radio.parent();
 			var containerSelector = $radio.attr('data-toggle');
 			var $containerToggle = $(containerSelector);
 
 			if(checked) {
+				// reset all items in group
+				this.resetGroup();
+
 				$radio.prop('checked', true);
 				$lbl.addClass('checked');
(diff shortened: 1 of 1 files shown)


interactivellama:
This may be the cause of #1287 too. "Checked radio on init only works if it is last"

interactivellama:
Also fixes #1270! Three bugs, one PR! w00t!

PR #1283, opened 2015-05-09
Tree select bug

interactivellama:
Definitely fixes #1137. May fix #1259.

- Refactor (combine majority of code) selectFolder and selectItem to re-use code.
- Add `clearSelected` method

There were several problems with this single select unit test:
- Mouse clicks were not being tested, only API methods
- Folder selectable tree needs to test items also.
- Folder selectable tree needs to test items with folders previously selected, and folders with items previously selected
- Markup for folder selectable tree did not include a `tree-item` template and therefore created only folders and did not create any items (hence 4 folders/items are returned instead of the folders/correct 8 items).
- The selectItem method was being used to select folders.

Does this fix your issue, @MJ-Vakili ?

PR #1312, opened 2015-05-21
Tree select bug for 3.7.x branch

swilliamset:
merged #1283 into master out of order. should have merged to 3.7.x first.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2358,7 +2358,8 @@
 
 			</div>
 			<div class="btn-panel">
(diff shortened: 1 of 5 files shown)


interactivellama:
There should be a third commit. "Make selectTreeNode DRY, add deselectAll method"

swilliamset:
this is better.

for future reference we should not `cherry-pick`. however we should restore the branch and create a new PR with the right target. i was making it to complicated.

interactivellama:
Sounds good. Yeah, my branch should still be there unmodified.

PR #1319, opened 2015-05-26
Simplify checkbox control, limit initialization to label tag

### LINKED ISSUE #1318: Checkbox refactor and limit initialize tag to label
Revert the revert of the checkbox refactor to make it simpler. See #1185 and #1282.

interactivellama:
Revert the revert of the checkbox refactor to make it simpler. See #1185 and #1282. Fixes #1318.

This commit adds back in the checkbox refactor by @mbeard. This is a breaking change. *This is not my work.*

This reverts commit c38ab67e5e5d29941cffecad37f226e67b2ee7be and needs to be an additional commit, since the original commits already exist.

### DIFF
--- a/index.html
+++ b/index.html
@@ -17,6 +17,7 @@
 	<script type="text/javascript">if(window.console && window.console.clear){ window.console.clear(); }</script>
 
 	<link href="index.css" rel="stylesheet" type="text/css">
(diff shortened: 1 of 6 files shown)


swilliamset:
Already reviewed when merged into master prematurely.

PR #1327, opened 2015-05-30
 [Wizard] Resolve bug #1322

### LINKED ISSUE #1322: [Wizard] Example bug
Go to page: http://getfuelux.com/javascript.html#wizard-examples-sample-methods
Click to `Next` button to last step. See that `Complete` button appear.
Move to start again. Click `jump to named step (Template)` button.
See that instead `Complete` button will be `Next` at last step

MJ-Vakili:
Resolve #1322 
and if step is string.

$('.wizard').wizard('selectedItem', {
    step: "4"
  });

### DIFF
--- a/js/wizard.js
+++ b/js/wizard.js
@@ -357,7 +357,7 @@
 			if (selectedItem) {
 				step = selectedItem.step || -1;
 				//allow selection of step by data-name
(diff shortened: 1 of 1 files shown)


swilliamset:
Nice catch! For those playing along and home, the `step` value needs to be converted because the `setState` function is expecting numbers https://github.com/MJ-Vakili/fuelux/blob/master/js/wizard.js#L157. Thx again @MJ-Vakili.

PR #1328, opened 2015-05-30
Add all release tasks to prompt menu

### LINKED ISSUE #1070: Release automation stuff
All this needs to be _one step_:

1. `git checkout master && git pull origin master`

1. `git checkout 3.x && git pull origin:3.x`

1. `git merge master`

1. `grunt release` or `grunt release --minor`
      * Updates the `package.json` and `bower.json` using semver. This will build the `dist` folder and update version numbers within markdown files. Visually check that `dist` file banners have the new version and commit. You will need `SAUCE_API_KEY.yml`.

1. `git add dist && git add *.md && git add *.json && git commit -m 'release x.x.x'`

1. Upload contents of `dist` folder to Fuel CDN server via FTP.

1. Push to `origin:3.x`.

1. Create tag locally. `git tag -a x.x.x -m 'vx.x.x'`. Publish tag to Github: `git push origin x.x.x`. The tag commit should exist in the major.x branch.

interactivellama:
- Git commands run from a menu
- Add grunt-shell dependency
- Ignores YAML file with upload credentials

TravisCI will fail, of course, because of the lack of `grunt-shell` module. 

Originally, I had tried multiple prompts, but the idea of branching decision trees doesn't seem to work very well with the grunt structure, so instead you run `grunt release` 4 or 5 times--which is as close to "Edsel-like" as I think we can get.

Decision tree branching is still not supported, but multiple prompts are now implemented with defaults that expect a release to be built from `master` and pushed to major release branch (`3.x`).

Fixes #1070.


Gruntfile.js
```diff
@@ -421,6 +444,144 @@ module.exports = function (grunt) {
 				}
 			}
 		},
+		shell: {
```
swilliamset: move shell options into separate module. Gruntfile.js is becoming excessively large.

Gruntfile.js
```diff
@@ -596,14 +756,75 @@ module.exports = function (grunt) {
...
-	grunt.registerTask('dorelease', '', function () {
-		if (typeof grunt.config('bump.increment') === 'undefined') {
+	grunt.registerTask('release', 'Select tasks to release a new version', ['prompt:releaseTasks', 'release-tasks']);
+
+	grunt.registerTask('release-tasks', function() {
+		var task = grunt.config('releaseTask');
+
+		grunt.config('banner', '<%= bannerRelease %>');
+
+		if (task !== 'exit') {
```
swilliamset: use prompt steps to handle conditions and task tree

Gruntfile.js
interactivellama: I agree. I was thinking about using https://github.com/creynders/load-grunt-configs

Gruntfile.js
```diff
@@ -421,6 +444,144 @@ module.exports = function (grunt) {
...
+		shell: {
+			// Compile release notes while waiting for tests to pass. Needs Ruby gem. 
+			// Install with: gem install github_changelog_generator
+			notes: {
+				command: 'github_changelog_generator --no-author --unreleased-only --compare-link'
+			},
+			mergeLocalMasterAndMinorBranch: {
+				command: function() {
+					var command = [
+						'git checkout master',
```
swilliamset: consider grunt-git instead of so much shell

swilliamset:
> ? What branch would like to pull from upstream and build your release in (probably 3.x)? 3.7.x

Can we list the available branches to release from?

swilliamset:
please make the remote repo configurable defaulting to `upstream` and listing the known remotes for the user to choose.

swilliamset:
@cmcculloh please take a look at this branch. your insight would be helpful.

interactivellama:
Updated with remote repo being configurable. 

As to listing remotes are you thinking of running a shell command and some kind of CLI text pipe that is then parsed by JS function within the prompt? The other option would be github API, but I'm not sure that's worth the effort considering you are 90% of the time going to pulling down master to build from

Gruntfile.js
```diff
@@ -355,19 +361,165 @@ module.exports = function (grunt) {
...
+								return answers['release.buildSemVerType'] === 'custom';
 							},
 							validate: function (value) {
 								var valid = semver.valid(value);
 								return valid || 'Must be a valid semver, such as 1.2.3-rc1. See http://semver.org/ for more details.';
 							}
+						},
+						{
+							config: 'release.remoteRepository',
+							default : 'upstream',
```
cormacmccarthy: Why are we repeating the defaults we set on line 44 here?

Gruntfile.js
```diff
@@ -355,19 +361,165 @@ module.exports = function (grunt) {
...
+							}
+						},
+						{
+							// Assumption is made that you are releasing the code within a "release branch" currently 
+							// on the upstream remote repo. This branch will be tracked locally and be used to run 
+							// the build process in. It will be named release_{BUILD_VERSION}_{MMSS} (that is, it will
+							// use the version specified earlier and a "mini-timestamp" of the current hour and minute).
+							config: 'release.remoteBaseBranch',
+							type: 'input',
+							default : '3.x',
```
cormacmccarthy: DRY

Gruntfile.js
interactivellama: Grunt-prompt defaults work if the user leaves the prompt blank. Even if there is a values set at the top in the `release` object, then an empty answer will override that and set it to empty string/undefined.

Gruntfile.js
interactivellama: The comment? Please elaborate. Are you asking for the two repo/branch prompts to merged and ask the person to write it with a slash?

swilliamset:
@interactivellama let me know when you've had a chance to push an update

Gruntfile.js
cormacmccarthy: What I'm saying is, why not use `grunt.config.release.remoteRepository` here instead of the string "upstream"?

So:
`default: grunt.config.release.remoteRepository,`

Gruntfile.js
cormacmccarthy: Don't Repeat Yourself. Same as my comment for line 378

interactivellama:
Updated with lo-dash compatible template strings. @cmcculloh @swilliamset

interactivellama:
Squashed and updated with master/3.8

interactivellama:
It only crashes the release task (and the saucelabs tests if those run and the `process.env` isn't present). I've added warnings to the top of the release task to warn releasers before the first prompt if those two files are not present.

Gruntfile.js
```diff
@@ -342,19 +348,165 @@ module.exports = function (grunt) {
...
+							// Assumption is made that you are releasing the code within a "release branch" currently 
+							// on the upstream remote repo. This branch will be tracked locally and be used to run 
+							// the build process in. It will be named release_{BUILD_VERSION}_{MMSS} (that is, it will
+							// use the version specified earlier and a "mini-timestamp" of the current hour and minute).
+							config: 'release.remoteBaseBranch',
+							type: 'input',
+							default : '<%= release.remoteBaseBranch %>',
+							message: function() {
+								return 'What remote branch from ' + grunt.config('release.remoteRepository') + 
+								' would like to build your release based upon?';
```
swilliamset: should be ' would _you_ like to build your release based on' ?

Gruntfile.js
```diff
@@ -408,6 +560,84 @@ module.exports = function (grunt) {
...
+			// Compile release notes while waiting for tests to pass. Needs Ruby gem and ONLY LOOKS AT THE REMOTE NAMED ORIGIN.
+			// Install with: gem install github_changelog_generator
+			notes: {
+				command: 'github_changelog_generator --no-author --unreleased-only --compare-link'
+			},
+			checkoutRemoteReleaseBranch: {
+				// this makes a local branch based on the prior prompt, such as release_{BUILD_VERSION}_{HOURSMINUTES}
+				command: function() {
+					grunt.config('release.localBranch', 'release_' + semver.inc(packageVersion, grunt.config('release.buildSemVerType')) + 
+							'_' + grunt.template.today('hMM'));
```
swilliamset: make this just a timestamp. hMM can still create dupes AM vs PM.

swilliamset:
next package versions in prompt and resulting package version in updated files come out differently. possibly using last tag versus package.json to determine options. needs clean up.

also `git fetch upstream --tag` should be ran as well so tags are up to date

Gruntfile.js
interactivellama: I'm [old fashioned](http://www.quora.com/What-is-the-difference-between-on-and-upon), I guess. 'On' is probably easier to understand in a publicly accessible repository for non-native English speakers.

interactivellama:
Updated to do track remote branch first. Then look at `package.json`and build the build prompt which should make it not matter what branch you start in as long as the `gruntFile` is up to date.

swilliamset:
Sorry for the confusion. Can we include the version and the timestamp in the temp release branch?

Also some coloring on the yml file warning would be good.

Otherwise this whole thing is a huge step forward and i'm accepting.

PR #1329, opened 2015-06-01
Add support for CommonJS, and publish to NPM

interactivellama:
We may promote use of the more feature rich npm module workflow in the future instead of bower. This at least opens the door to that possible effort. This does not however make the library compatible with CommonJS, yet though.


vernak2539:
I feel this should be done all at once. If you do this, you can just publish it, that's it. It won't really work, since nothing has been done to enable it for CJS.

I spent a few hours going over what needs to be done to get it working.

1. bootstrap and jquery need to be added as dependencies (should also remove server-static and connect since they are in the wrong spot if this is going on npm).
    * jquery@1.9.1 does not support CJS
1. in every js file, a if else needs to be added to check for module and module.exports, then actually use the factory with "node style" requiring of jquery
1. a main file needs to be added to the package.json
   * should probably also create a dist npm file that requires everything instead of trying to use the ./js/all.js for both browser, AMD, and CJS
1. probably a few more things....

Was going to try and work on this sooner than later, but above is the general gist to getting it working

interactivellama:
Thanks Alex for delimiting these. I agree it might piss some people off to publish but not support commonJS. I'll modify this PR with the above suggestions.

interactivellama:
@vernak2539 It appears that even though Bootstrap "supports" CommonJS packaging, it still is asking for jQuery on the window object. In Fuel UX, AMD users can get around this with requirejs shim in their config object.

Do we need to to wrap BS somehow ourselves like we suggest with AMD or is it OK to depend on the global for jQuery? I am kind of new to CommonJS on the front-end space, so I wanted to see if you had any common usage suggestions.

interactivellama:
@vernak2539 I have a [WIP branch](https://github.com/ExactTarget/fuelux/commit/24a1ab41ec582b5078f931e793a4024cc1a30906) here for you to test. I didn't want to force push it here, because it's got a lotta cruft in it that's not going in the library in order to just use it (and I didn't want to scare anyone into thinking it was all going in). It has Fuel UX controls running as commonJS modules in `/index-cjs.html`.

I _don't_ think it would be good idea to duplicate all the tests just in a commonJS format.

vernak2539:
@interactivellama yeah ill take a look today. Definitely should be no need to duplicate the tests.

interactivellama:
One thought I had is that I didn't add the CommonJS wrapper to `dist/js/fuelux.js` would anyone expect that to be present? I would think they would always use the the `/js` via node_modules, since "small is good" and "combined is bad" for CommonJS.

vernak2539:
I think that looks fine. You'll probably have to have a gist or something demonstrating how it needs to be setup with jquery, bootstrap, etc.

not sure what you should do about including bootstrap. usually in these kind of things, it would be stated as a dependency, and thus installed on the npm install. But, this is a repo for all the various types. May want to look into that, and making sure only real dependencies are listed in the package.json

vernak2539:
>One thought I had is that I didn't add the CommonJS wrapper to dist/js/fuelux.js would anyone expect that to be present? I would think they would always use the the /js via node_modules, since "small is good" and "combined is bad" for CommonJS.

Since you've set the "main" script in the package.json, all that's required is `require('fuelux')` and it should sort the correct package. If people use a relative path when trying to access it, they're using it wrong

interactivellama:
Looks like I forget to push the [latest updates](https://github.com/interactivellama/fuelux/blob/6eed5126140fa9d4c68b37400e1d06568692a114/package.json) to package.json that add jquery and bootstrap as dependencies.

vernak2539:
>Looks like I forget to push the latest updates to package.json that add jquery and bootstrap as dependencies.
yep that looks good

interactivellama:
Updated with actual code to support commonJS + basic unit tests ran within the `validate-dist` grunt task.

interactivellama:
@swilliamset Updated with master branch. Also tests finally passed by getting `grunt-browserify`on the heroku app.

PR #1331, opened 2015-06-02
bug-fix/npm scripts for windows

vernak2539:
The postInstall cmd did not work on windows, this fixes that.

The tar command didn't work as well b/c some modules have file paths too long for windows to comprehend. Since the tar is ignored form the repo and only used in the travis build, I moved that command to be a step before npm test (only run on travis I assume).

If this is not the correct way, please let me know as this problem has more of a barrier to entry for devs on windows

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,8 +70,8 @@
   },
   "scripts": {
     "main": "devserver.js",
-    "postinstall": "./node_modules/bower/bin/bower install && tar -zcf dev/node_modules.tar.gz node_modules/;",
-    "test": "grunt travisci --verbose"
+    "postinstall": "node ./node_modules/bower/bin/bower install",
+    "test": "tar -zcf dev/node_modules.tar.gz node_modules/ && grunt travisci --verbose"
   },
   "title": "Fuel UX",
   "version": "3.7.3",


vernak2539:
I'm actually not sure why ./dev/node_modules.tar.gz is there. the travis test pulls it from the fuelux heroku app, I only assume this has been done to limit the fact if the registry goes down, but doesn't seem necessary as the registry has been stable for a while

vernak2539:
I guess what I'm getting at is the fact that the tar command is not in the correct place if it's actually needed. Trying to work on some stuff, but every npm install generates an error log on windows

interactivellama:
Travis doesn't need to run the tarball. Only the "edge server" heroku app does. TravisCI downloads the modules from there. Having a Heroku-only hook would be the best option. Maybe a bash script that asks if it's on Heroku, etc and then runs the tarball?

The best option may be to set an ENV variable that is only present on the edge server and the tarball only runs if it is present. Like `MAKE_FUELUX_TARBALL=1`

interactivellama:
`MAKE_FUELUX_TARBALL=1` is now present on the [edge server](https://fuelux-dev.herokuapp.com/).

interactivellama:
Now that I think of it, the edge has to grunt in development mode (as it currently is), so moving this to a grunt task would be fine.

interactivellama:
Will keep open until I PR a fix.

vernak2539:
so for the windows working, do I need another PR or can you change that in the new one you opened?

interactivellama:
My PR should fix it. I left `bower install` in there, but moved the `tar` command to the Procfile so that it only runs on Heroku.

No one else needs to run that `tar` command that isn't administrating this library.

vernak2539:
okay, commented on the PR of where you need to change it still

PR #1333, opened 2015-06-02
Make install process Windows compatible

interactivellama:
Move node_module archiving to Procfile `tar` command, so it only run on Heroku, and fixes #1331 for those that run on Windows or don't have `tar` installed. It also adds `node` before the bower command which makes it Windows compatible.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: grunt dist && node devserver.js
+web: tar -zcf dev/node_modules.tar.gz node_modules/ && grunt dist && node devserver.js
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
   },
   "scripts": {
     "main": "devserver.js",
-    "postinstall": "./node_modules/bower/bin/bower install && tar -zcf dev/node_modules.tar.gz node_modules/;",
+    "postinstall": "node ./node_modules/bower/bin/bower install",
     "test": "grunt travisci --verbose"
   },
   "title": "Fuel UX",


package.json
```diff
@@ -70,7 +70,7 @@
   },
   "scripts": {
     "main": "devserver.js",
-    "postinstall": "./node_modules/bower/bin/bower install && tar -zcf dev/node_modules.tar.gz node_modules/;",
+    "postinstall": "./node_modules/bower/bin/bower install",
```
vernak2539: for it to work on windows this needs to be 

```
"postinstall": "node ./node_modules/bower/bin/bower install",
```

interactivellama:
Updated @vernak2539 . I'm sorry I didn't understand the comment.

vernak2539:
:+1:

Procfile
```diff
@@ -1 +1 @@
-web: grunt dist && node devserver.js
\ No newline at end of file
+web: tar -zcf dev/node_modules.tar.gz node_modules/ && grunt dist && node devserver.js
```
swilliamset: The tar step is to insure the node_modules in the dev environment deploy quickly to heroku correct? Or is there more going on here?

interactivellama:
The archive is created in order to allow TravisCI to not run `npm install`. 

It's what Bootstrap does, except they use Python and AWS.

swilliamset:
right on

vernak2539:
:+1:

PR #1334, opened 2015-06-02
Removes 2-step build process, improve CSS/LESS debugging

### LINKED ISSUE #1227: Update LESS build to have the `.fuelux` wrapper without the two-stage build.
As was discovered in mctheme, we can have the best of both worlds: the `&` parent selector AND a wrapper class, all without the hacky "namespace" two-stage build we are currently doing.

Will drastically improve developer happiness because source maps and less confusion.

interactivellama:
Add fuelux class across LESS files. This is less DRY, but allows LESS parent selectors (`&`) to work properly without a two step build process like we currently have. This also also CSS source maps to work better allow debugging to actual line numbers in the source LESS. Fixes #1227.


vernak2539:
Just wondering, is there a reason you put did it like this for all of them

```css
.fuelux .checkbox { ....
```

when  you could do something more along the lines of LESS syntax like
```less
.fuelux {
    .checkbox {
        .....
```
All files that were changed in this PR  in multiple places would be now only changed in one spot

interactivellama:
@vernak2539 I was taking my pattern from what was completed in MC theme which was line by line, but yes, after some testing it does look like the majority of the LESS files in this repo could be wrapped. `form.less` is probably the heaviest user of the pseudo parent selector (`&`).

interactivellama:
Updated. Ignore white space: https://github.com/ExactTarget/fuelux/pull/1334/files?w=1

vernak2539:
@interactivellama looks good to me. a little cleaner also I feel

swilliamset:
extra line breaks are inconsistent. sometimes there is one after the `.fuelux {` line sometimes there isn't. be consistent por favor.

PR #1339, opened 2015-06-08
Remove startDate overwrite

### LINKED ISSUE #1293: Scheduler Sets Start dateTime to the Reoccuring Until Date  no matter what
The Scheduler control in version 3.7.1+ sets its StartDateTime datepicker control to the Reoccuring Until DateTime if you initialize scheduler with a StartDateTime, TimeZone, and ReoccurancePattern. You should remove line 6999 and then change line 7000 to use the 'temp' variable instead of the startDate variable. If you don't do this you'll never be able to initialize the start datetime with a different datetime than the Reoccurance Until date. The lines numbers are from version 3.7.1 however the same problem is in 3.7.2

swilliamset:
could go into 3.7.4 if need be or 3.8.0 (whichever goes first)

Fixes #1293.

### DIFF
--- a/js/scheduler.js
+++ b/js/scheduler.js
@@ -580,8 +580,7 @@
 					var timeZone = this.$timeZone.selectlist('selectedItem');
 					var timezoneOffset = (timeZone.offset === '+00:00') ? 'Z' : timeZone.offset;
 
-					startDate = temp;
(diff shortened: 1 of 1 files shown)


interactivellama:
Thanks @razor247!

Issue #1267, opened 2015-04-27
Enable Frozen Columns on right side


BenjaminNeilDavis:
Currently only supports left side columns. Create option that would allow for both left and right columns to be frozen.

Issue #1311, opened 2015-05-21
native support of checkbox in multi select repeater


swilliamset:
* single select
  * highlighted row 
  * checkmark
* multi select
  * check box
  * no highlight
  * frozen left column

PR #1340, opened 2015-06-08
Created item actions frozen column on right side

BenjaminNeilDavis:
Fixes issue #1267 
- Creates new frozen actions column if enabled
- there will be two follow up pull requests
- one to solve this issue https://github.com/ExactTarget/fuelux/issues/1311 and one to integrate this with actions / frozen columns

ex:
```javascript
list_actions:  {
						width: 50,
						items: [
							{
								name: 'edit',
								html: function () {
									return '<div class="fuelux-icon fuelux-icon-pencil"></div> Edit'
								}
							},
							{
								name: 'copy',
								html: function () {
									return '<div class="fuelux-icon fuelux-icon-copy"></div> Copy'
								}
							},
							{
								name: 'delete',
								html: function () {
									return '<div class="fuelux-icon fuelux-icon-delete"></div> Delete'
								},
								clickAction: function(helpers, callback) {
									console.log('hey it worked');
									console.log(helpers);
									callback();
								}
							}
						]
					},
```

### DIFF
--- a/js/repeater-list.js
+++ b/js/repeater-list.js
@@ -263,9 +263,9 @@
 
 				this.$canvas.addClass('actions-enabled');
 			}
-			this.$element.find('.repeater-list table.table-actions tr').each(function (i, elem) {
-				$(this).height($table.find('tr:eq(' + i + ')').height());
-			});
+
+			this.$element.find('.repeater-list .actions-column-wrapper, .repeater-list .actions-column-wrapper td, .repeater-list .actions-column-wrapper th')
+				.css('width', this.list_actions_width);
 
 			this.$element.find('.table-actions .action-item').on('click', function() {
 				var actionName = $(this).data('action');
@@ -325,6 +325,7 @@
 			initialize: function (helpers, callback) {
 				this.list_sortDirection = null;
 				this.list_sortProperty = null;
+				this.list_actions_width = (this.viewOptions.list_actions.width !== undefined) ? this.viewOptions.list_actions.width : '37px';
 				callback();
 			},
 			resize: function () {
@@ -415,7 +416,7 @@
 
 		var property = columns[columnIndex].property;
 		if(this.viewOptions.list_actions !== false && property === '@_ACTIONS_@'){
-			content = '<div class="repeater-list-actions-placeholder" style="width: ' + this.viewOptions.list_actions.width  + '"></div>';
+			content = '<div class="repeater-list-actions-placeholder" style="width: ' + this.list_actions_width  + '"></div>';
 		}
 
 		content = (content!==undefined) ? content : '';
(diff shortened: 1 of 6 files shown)


kevinparkerson:
Noticing the rows being off by about a pixel in Firefox, but not in Chrome. Will keep investigating

js/repeater-list.js
```diff
@@ -201,7 +300,8 @@
 			list_selectable: false,
 			list_sortClearing: false,
 			list_rowRendered: null,
-			list_frozenColumns: 0
+			list_frozenColumns: 0,
+			list_itemActions: false
```
kevinparkerson: I think we should call this option `list_actions` to simplify (replacing the name throughout the control). Then, I'd suggest making this accept an object vs an array. The object would have an `items` attribute, which would be the array of action items :) The reason for the object is I think we need other attributes here, such as `label` and `width`. Since the actions column, while fixed, is essentially an additional column, we need a means to set it's column label and width just like we have via the columns settings.

kevinparkerson:
We will essentially need to create an automatic additional column for actions if `list_actions` is set to an object. The reasoning for this is that if you have lots of content in the last column, it will be obscured by the actions column (try putting a paragraph of lorem ipsum for content in last column of the incoming data to see this currently). As such, the actions column needs to be an actual column on the repeater-list, with the fixed table positioned above it so that it does not scroll (similar to the floating columns on the left).

The settings for the additional column should be as follows:
```
{
	label: this.viewOptions.list_actions.label || '',
	property: '',
	sortable: false,
	width: (this.viewOptions.list_actions.width !== undefined) ? this.viewOptions.list_actions.width : {{defaultActionsWidth}}
}
```
This additional column should be added in `renderThead`. To prevent the `differentColumns` function from acknowledging this dynamically added column, on line 606 of `renderThead` check if `list_actions` is not equal to `false`, and if that is the case remove the last item from `this.list_columns` (which would be the dynamically added action column). Then, after line 610 where `this.list_columns` is set to `columns`, if `list_actions` is set to false add the above object to the end of `this.list_columns`. That should make the appropriate actions column so that the last column never goes beneath it. Because property is set to `''`, in `renderColumn` there is one other change needed before line 414: `content = (content!==undefined) ? content : '';` This is to prevent unexpected results from adding content to a column, because for actions the property will not be found in the column definition and as such when it looks up the content for that particular item it will come back as `undefined`.

kevinparkerson:
Note: I modified my suggestions for the dynamic column in `renderThead` after realizing a potential issue.

cormacmccarthy:
Needs:

* unit tests
* dev examples (in /index.html and markup/)
* documentation

BenjaminNeilDavis:
@cormacmccarthy That is correct working on all of that now thanks!

kevinparkerson:
@BenjaminNeilDavis Code looks good, still seeing some stylistic issues in Firefox involving the action column rows not aligning with the rest of the row. Might need to change that `specialBrowserClass` function to return a string of classes such as `b-{{browser}} v-{{browserVersion}}`, that way you could add those to the repeater wrapper and target Firefox specifically if you need to via `b-firefox`

also noticing an issue in Chrome where the actions column has a left border that is appearing above the `repeater-list-heading`

PR #1351, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -25,7 +25,7 @@ module.exports = function (grunt) {
 		' */\n',
 		bump: {
 			options: {
-				files: ['bower.json', 'package.json'],
+				files: ['package.json'],
 				updateConfigs: ['pkg'],
 				commit: false,
 				createTag: false,
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,6 @@
 {
   "name": "fuelux",
   "description": "Extending Bootstrap with additional lightweight JavaScript controls.",
-  "version": "3.8.0",
   "keywords": [
     "application",
     "bootstrap",


vernak2539:
This is required for the "bower badge" in the readme file. will not remove

vernak2539:
Actually,

> Version badge uses GitHub tags to identify the latest Bower version. Please use Semantic Version tags to ensure proper badge function.

This is valid. not sure the bower version is used for anything

kkirsche:
It's not, hence the PR. Also just an FYI there are discussions to move bower to NPM anyway so I wouldn't really plan to support it for ever: bower/bower#1520

kkirsche:
Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

interactivellama:
We modify the `bower.json` file via the `replace` task in `gruntFile.js` at every release, so that task needs to be modified too.

swilliamset:
The following line will need to be updated to no longer attempt to bump the version in the bower file

https://github.com/ExactTarget/fuelux/blob/master/Gruntfile.js#L28

@kkirsche would you be willing to update the PR

kkirsche:
@swilliamset done

swilliamset:
Looks like we have a conflict now. Could you rebase? If not totally understand we'll handle it. Thanks for the heads up.

kkirsche:
@swilliamset yeah it was cause you guys issued another release since this was made. I've rebased and squashed cause, well, why not

swilliamset:
squashes are always greatly appreciated. once travis finishes i'll be glad to merge. thanks again.

PR #1352, opened 2015-06-11
Wrap wizard steps in hidden overflow instead of entire control

### LINKED ISSUE #1241: Make wizard wrapper `div` overflow visible
Is there any way how to set overflow: visible to Fuelux wizard top level div so the bigger content is shown correctly and wizard orientation header is shown correctly (it flows outside the wizard boarder in this case). FuelUX wizard obviously counts with overflow-x: hidden since it hides individual steps when moving across them.

Than the only option is to set overflow-x: hidden and overflow-y: auto (since it is the option supported by browsers - hidden + visible is not). But this looks really ugly then.

Simple scenario: I have wizard page with longer combobox in it.

kevinparkerson:
Fixes #1241 by adding additional wrapper element with class `.steps-container` around the `ul.steps`, so that `overflow: hidden` is only applied to that area instead of the entire `wizard`.

This is a breaking markup change that maintains backwards compatibility by adding the class `.no-steps-container` to `.wizard` via JavaScript. Please add the `div.steps-container` element around `ul.steps` as [this markup documentation](http://getfuelux.com/javascript.html#wizard-usage-markup) when you can.

for easier comparison 
https://github.com/ExactTarget/fuelux/pull/1352/files?w=1

### DIFF
--- a/index.html
+++ b/index.html
@@ -2373,18 +2373,20 @@
 				<h2>Wizard</h2>
 				<div class="thin-box">
 					<div class="wizard" data-initialize="wizard" id="myWizard">
-						<ul class="steps">
-							<li data-step="1"><span class="badge">1</span>Campaign<span class="chevron"></span>
-							</li>
-							<li class="active" data-step="2"><span class="badge">2</span>Recipients<span class="chevron"></span>
-							</li>
-							<li data-step="3"><span class="badge">3</span>Template<span class="chevron"></span>
-							</li>
-							<li data-step="4"><span class="badge">4</span>Preview<span class="chevron"></span>
-							</li>
-							<li data-step="5" data-name="distep"><span class="badge">5</span>Send<span class="chevron"></span>
-							</li>
-						</ul>
+						<div class="steps-container">
+							<ul class="steps">
+								<li data-step="1"><span class="badge">1</span>Campaign<span class="chevron"></span>
+								</li>
+								<li class="active" data-step="2"><span class="badge">2</span>Recipients<span class="chevron"></span>
+								</li>
+								<li data-step="3"><span class="badge">3</span>Template<span class="chevron"></span>
+								</li>
+								<li data-step="4"><span class="badge">4</span>Preview<span class="chevron"></span>
+								</li>
+								<li data-step="5" data-name="distep"><span class="badge">5</span>Send<span class="chevron"></span>
+								</li>
+							</ul>
+						</div>
 						<div class="actions">
 							<button type="button" class="btn btn-default btn-prev"><span class="glyphicon glyphicon-arrow-left"></span>Prev</button>
 							<button type="button" class="btn btn-default btn-next" data-last="Complete">Next<span class="glyphicon glyphicon-arrow-right"></span>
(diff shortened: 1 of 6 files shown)


index.html
```diff
@@ -2403,7 +2405,7 @@ <h2 class="header">Year</h2>
 								<p>Pea horseradish azuki bean lettuce avocado asparagus okra. Kohlrabi radish okra azuki bean corn fava bean mustard tigernut jÃ­cama green bean celtuce collard greens avocado quandong fennel gumbo black-eyed pea. Grape silver beet watercress potato tigernut corn groundnut. Chickweed okra pea winter purslane coriander yarrow sweet pepper radish garlic brussels sprout groundnut summer purslane earthnut pea tomato spring onion azuki bean gourd. Gumbo kakadu plum komatsuna black-eyed pea green bean zucchini gourd winter purslane silver beet rock melon radish asparagus spinach. </p>
 								<p>Beetroot water spinach okra water chestnut ricebean pea catsear courgette summer purslane. Water spinach arugula pea tatsoi aubergine spring onion bush tomato kale radicchio turnip chicory salsify pea sprouts fava bean. Dandelion zucchini burdock yarrow chickpea dandelion sorrel courgette turnip greens. </p>
 							</div>
-							<div class="step-pane sample-pane alert-danger alert" data-step="3">
+							<div class="step-pane sample-pane alert-danger alert" data-step="3" style="position: relative;">
```
swilliamset: Move inline style to less

index.html
```diff
@@ -2373,18 +2373,20 @@ <h2 class="header">Year</h2>
...
-							<li class="active" data-step="2"><span class="badge">2</span>Recipients<span class="chevron"></span>
-							</li>
-							<li data-step="3"><span class="badge">3</span>Template<span class="chevron"></span>
-							</li>
-							<li data-step="4"><span class="badge">4</span>Preview<span class="chevron"></span>
-							</li>
-							<li data-step="5" data-name="distep"><span class="badge">5</span>Send<span class="chevron"></span>
-							</li>
-						</ul>
+						<div class="steps">
```
swilliamset: instead of moving steps class (causing js selectors to need to be update) please add a new class

less/wizard.less
```diff
@@ -10,178 +10,185 @@
...
 
-			&.previous-disabled {
-				li {
-					&.complete {
-						cursor: default;
+			ul {
+				list-style: none outside none;
+				padding: 0;
+				margin: 0;
+				width: 999999px;
```
swilliamset: magic number. please include a comment explaining why.

index.html
kevinparkerson: Ah, this was me testing that overflow was no longer clipped via some quick inline hackery and forgetting to remove it once I was done.

less/wizard.less
kevinparkerson: that was `width: 4000px;` before. I bumped this number to `999999` to give it a ridiculous amount of space for the li elements to flow horizontally

interactivellama:
Looks good. Please squash.

PR #1357, opened 2015-06-17
If selectlist is empty, disable it

### LINKED ISSUE #1349: An empty selectlist needs to not look like a scrunched-up dropdown.
An empty selectlist should still have a min-height and min-width so that it is not too visually small. Currently if no items are present the overall size is approximately 42 pixels in width by 14 pixels in height. The min-height should probably be the standard height of the selectlist if items are present and the min-width perhaps twice that of the min-height.

interactivellama:
Fixes #1349 by checking to see if `li` are present and disabling itself if there are none or only one single item.

What's the point of selectlist with 1 item or no items?

Also implements an empty list label API:

```
var $selectlist = $(html).find('#selectlistEmpty').selectlist({
	emptyLabelHTML: '<li data-value=""><a href="#">I am feeling so empty</a></li>'
});
```

This is getting a little imperative with the `emptyLabelHTML`.

**UPDATED:** Design prefers to behave as a native `select` element and still allow the dropdown if only a single item exists.

### DIFF
--- a/js/selectlist.js
+++ b/js/selectlist.js
@@ -31,6 +31,7 @@
 		this.$element = $(element);
 		this.options = $.extend({}, $.fn.selectlist.defaults, options);
 
+
 		this.$button = this.$element.find('.btn.dropdown-toggle');
 		this.$hiddenField = this.$element.find('.hidden-field');
 		this.$label = this.$element.find('.selected-label');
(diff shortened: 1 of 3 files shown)


jamin-hall:
image

@interactivellama let's try it this way.

interactivellama:
How do we want to allow the developer to create "a blank selectlist"? Right now, you can create the state by adding the empty option to the dropdown, initializing it, and then disabling it. Do we want to make this easier?

kevinparkerson:
We probably want a min-height on the selectlist, to ensure blank selectlists are not super tiny as well

PR #1358, opened 2015-06-17
If combobox dropdown is empty, disable the dropdown only

### LINKED ISSUE #1349: An empty selectlist needs to not look like a scrunched-up dropdown.

interactivellama:
Fixes #1349 by checking to see if `li` are present and disabling dropdown only if there are none. Input still works.

### DIFF
--- a/js/combobox.js
+++ b/js/combobox.js
@@ -43,6 +43,13 @@
 
 		// set default selection
 		this.setDefaultSelection();
+
+		// if dropdown is empty, disable it
+		var items = this.$dropMenu.children('li');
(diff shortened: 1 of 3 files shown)


jamin-hall:
@interactivellama :+1:

kevinparkerson:
Good stuff!

PR #1360, opened 2015-06-18
Gray out disabled checkbox/radio while maintaining accessibility

### LINKED ISSUE #1380: Add WCAG approved opacity to checkbox labels

interactivellama:
Make disabled radio/checkbox and their labels opacity 65% instead of 50% opacity.

WCAG AA: Pass
WCAG AAA: Pass (borderline)

http://www.w3.org/WAI/WCAG1AAA-Conformance

@jamin-hall I'm getting a ratio around 7:1 which should be AAA.

Fixes #1380.

### DIFF
--- a/less/checkbox.less
+++ b/less/checkbox.less
@@ -90,11 +90,12 @@
 
 		&.disabled {
 			cursor: not-allowed;
+			opacity: .65;
 
 			&:before
 			{
 				cursor: not-allowed;
-				opacity: .5;
+				opacity: .65;
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


jamin-hall:
@interactivellama let's go with 65% opacity, but we'll need to update both the "no-javascript" and "javascript" versions to be consistent. In the future we should consider AAA, but at the very least we should meet AA standards.

kevinparkerson:
Always good to provide accessibility! The opacity change seems to be good here... on a sort-of-related note, I'm noticing in both Firefox and Chrome that tabbing does not appear to be working correctly. When I click the start of the page and hit tab a bunch, it appears to be selecting the checkboxes as it takes quite some time to get to the checkbox example method buttons, but no indicator is shown. @interactivellama are you seeing this? I recall having some indicator upon tabbing before...

interactivellama:
#1382 fixes that issue.

kevinparkerson:
Cool, new issue for that. This one is good then!

PR #1363, opened 2015-06-19
Focus on first matching item when key is pressed

### LINKED ISSUE #128: Focus on first matching item in selectlist dropdown when key is pressed
It's exactly like a normal <select> works. 
Once you click it and the options appears if you type, you jump directly to the proper result.

interactivellama:
Fixes #128

I've made the mistake of thinking I could jump down to the next item when I pressed a key before in a custom selectlist. This allows a feature that the native `select` supports. When the dropdown is open pressing a letter will search the dropdown list until the first item starting with that letter is found and set the focus to that `a` tag. 

@jamin-hall Do we want to support this?

### DIFF
--- a/js/selectlist.js
+++ b/js/selectlist.js
@@ -42,6 +42,30 @@
 		if (options.resize === 'auto' || this.$element.attr('data-resize') === 'auto') {
 			this.resize();
 		}
+
+		// support jumping focus to first letter in dropdown when key is pressed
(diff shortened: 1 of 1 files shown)


jamin-hall:
@interactivellama, yep. This sounds great.

kevinparkerson:
tested this and it's awesome... nice add!

PR #1369, opened 2015-06-23
Prevent selectlist disabled items to be selected

### LINKED ISSUE #1232: Add support for disabled menu items in selectlist
We should add support for disabled states within selectlist's dropdown menus when adding the `.disabled` class.

See [Bootstrap documentation](http://getbootstrap.com/components/#disabled-and-active-states) and PR #1181 into the Fuel UX 2 branch for more information.

interactivellama:
Fixes #1232.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2377,6 +2377,8 @@
 							</li>
 							<li data-value="Item Five" data-foo="bar" data-fizz="buzz"><a href="#">Item Five</a>
 							</li>
+							<li class="disabled" disabled="disabled" data-value="disabled"><a href="#">Disabled item</a>
+							</li>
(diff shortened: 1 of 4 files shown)


kevinparkerson:
@interactivellama do we want to prevent the `doSelect` function from selecting disabled items as well? That way no selection APIs can do this to an explicitly disabled item. I'm not sure we should prevent the API from doing so, but worth considering.

kevinparkerson:
@interactivellama you know what, never mind... I think such a restriction would go against the flexibility we allow with our APIs in other controls. This is a user restriction, not a programmer restriction. I'm going to merge this :)

PR #1371, opened 2015-06-24
Removes trailing semicolon from scheduler recurrance pattern

### LINKED ISSUE #1370: Scheduler reccurance returns a pattern ending in a semi-colon
Although probably not doing any harm. The standard says that semicolons should only be used to separate.

interactivellama:
Fixes #1370. However, I am unsure if this a breaking change for anyone if the semicolon is removed. I'm estimating that any library that is parsing ical recurrance pattern is using a `split(";")` method and so this doesn't matter. Our unit tests however needed changing.

### DIFF
--- a/index.js
+++ b/index.js
@@ -762,7 +762,7 @@ define(function (require) {
 				"name": "Namibia Standard Time",
 				"offset": "+02:00"
 			},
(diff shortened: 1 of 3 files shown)


kevinparkerson:
@interactivellama this change looks good. I don't think it should be a big issue, but it is something that should be put in the release notes for sure

PR #1372, opened 2015-06-24
Allow empty string (ISO 8601) in datepicker momentJS format

interactivellama:
Format can be an empty string because that is the API for the ISO 8601 format in MomentJS.


kevinparkerson:
looks good!

PR #1373, opened 2015-06-25
Correct scheduler selection of `.year-month`

### LINKED ISSUE #1307: Schedular yearly dropdown of month days always returns 1.
There is a jquery selector problem in the yearly and monthly controls. They are not specific enough so your selectors are returning two instances of the controls and failing. This is why you always get a 1 as a return from the dropdown no matter what one you pick. You selectors need to start with "Yearly <control ClassName> and Monthly <ControlClassName> instead of just <ControlClassName> because the controls are similarly named and return two controls instead of one without denoting the parent.

interactivellama:
Increases element selection specificity and fixes #1307.


kevinparkerson:
good catch

PR #1374, opened 2015-06-25
Disable repeater pagination if 0 or 1 pages returned

interactivellama:
screen shot 2015-06-25 at 12 31 24 pm
screen shot 2015-06-25 at 12 31 04 pm


kevinparkerson:
looks good!

Issue #972, opened 2015-01-10
Don't override bootstrap background colours to allow bootstrap themes to work with fuelux


TMiguelT:
I've been developing an app using the [superhero bootswatch theme ](http://bootswatch.com/superhero/), which is a dark theme that uses light text (unlike the default bootstrap theme). The problem is that some of the fuelux widgets (primarily *repeater* and *wizard*) have hard coded white backgrounds, which means they look terrible when combined with a dark theme, and worst of all the text is illegible.

 I added superhero to your [demo page](http://getfuelux.com/javascript.html) as a demonstration. You can do the same to see what I mean:
```javascript
var l = document.createElement("link");
l.rel = "stylesheet";
l.href = "//maxcdn.bootstrapcdn.com/bootswatch/3.3.0/superhero/bootstrap.min.css";
b = document.querySelector("body");
b.appendChild(l);
```
Examples:
![capture](https://cloud.githubusercontent.com/assets/5019367/5691649/b905828c-9926-11e4-98df-85ba6b1f627e.PNG)
![capture](https://cloud.githubusercontent.com/assets/5019367/5691651/cb4ea018-9926-11e4-9251-2ad0a3dcb30d.PNG)

Would it be possible to use standard bootstrap classes like `.panel` and `.container` to make fuelux work better alongside bootstrap themes? Thanks so much for the library!

interactivellama:
Thanks @TMiguelT. This is something to consider. We've had to theme these controls manually ourselves, too. See [Themed Fuel UX Controls](http://getfuelux.com/mctheme/index.html#mctheme) at the bottom. [These are the overrides](https://github.com/ExactTarget/fuelux-mctheme/tree/master/less/fuelux-override), we've had to write ourselves.

interactivellama:
@TMiguelT How do other Bootstrap extensions (JS plugins) handle this? Are you using any that are as complex that handle this well?

TMiguelT:
Well all bootstrap plugins I've used use bootstrap classes so I've never noticed this problem before. I did a little playing around, and I think with very little effort the wizard could be fixed to work with custom themes.

In the below examples I only changed 3 things with the wizard CSS:
* Added the `.panel` class to the `.wizard` element
* Removed the `background-color` from `.fuelux .wizard`
* Removed the `background-color` from `.fuelux .wizard > .actions` (the top right panel with the buttons)

If you compare them to the examples I posted at the top, I think they're a lot better. They're certainly legible now.

[...]

PR #1379, opened 2015-06-29
Color consolidation into variables

### LINKED ISSUE #856: Less variable refactoring (bring layout pixels and color hex values in variables)
I would like to use fuelux in my theme, but the less files don't use variable.less.
For example you define colors inside checkbox.less. Sadly there is not much point in using variable.less this way...

Are you planning to fix this?

interactivellama:
Removes all hex values from LESS files. There were a few grays that had minimal saturation that were converted to hue-less grays. A few grays may have changed by 1%-3% luminance in order to create consistency. There are still 20+ grays being used.

Fixes #856 and makes getting to the idea of #972 easier although developers will still need to modify `fuelux/less/variables.less` if they would like to theme the controls.

### DIFF
--- a/less/checkbox.less
+++ b/less/checkbox.less
@@ -16,7 +16,7 @@
 			}
 
 			&.checked label.checkbox-custom, label.checked.checkbox-custom {
-				background: #e9e9e9;
+				background: @gray90;
 				border-radius: @baseBorderRadius;
 			}
 		}
@@ -26,7 +26,7 @@
 		input[type="checkbox"]:hover + .checkbox-label,
 		.checkbox-label:hover
 		 {
-			color: @grayLight;
+			color: @gray60;
 		}
 
 		input[type="checkbox"]:disabled:focus + .checkbox-label,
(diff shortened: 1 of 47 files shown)


interactivellama:
Removed `dist` files and squashed!

less/checkbox.less
```diff
@@ -51,10 +51,10 @@
 		position: relative;
 
 		&:before {
-			#fuelux > .icon-base-styles("Glyphicons Halflings");
```
swilliamset: this branch needs to be rebased/merged with the mixin namespacing solution

interactivellama:
Updated. `grunt-browserify` was not on the Heroku edge server.

interactivellama:
@swilliamset Updated by removing edits from other PRs that are now present since merging the past few days.

swilliamset:
We need to consider whether or not `lighten()` and `desaturate()` against a variable is recommended. Should the resulting values be equal to a variable already? Otherwise excepting the PR.

PR #1382, opened 2015-06-30
Fix checkbox markup in example page to have span

### LINKED ISSUE #1381: Checkbox has no style change on focus
There is a style change on hover, but not focus.

interactivellama:
The `index.html` checkbox markup was incorrect. Fixes #1381.

### DIFF
--- a/index.html
+++ b/index.html
@@ -55,44 +55,49 @@
 
 					<section id="checkboxes-block">
 
-						<div class="checkbox">
+						<div class="checkbox" id="myCheckbox">
 							<label class="checkbox-custom" data-initialize="checkbox" id="myCustomCheckbox1">
-								<input type="checkbox" class="sr-only">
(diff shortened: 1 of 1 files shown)


kevinparkerson:
@interactivellama the grey-ification on tabbing is back! although it's somewhat hard to see... perhaps we add the hover box-shadow effect too? (if possible)

interactivellama:
The issue is that the input has the focus and it is a child of the label. There is no CSS parent selector. The hover is able to affect the input, but the input cannot affect the label, just the span which is its sibling.

swilliamset:
Is this resolved and ready to be merged or on hold?

interactivellama:
@swilliamset This is a revert to a previous state that got overwritten. The hover suggestion by @kevinparkerson is not possible with CSS only due to the lack of a reverse sibling selector.

PR #1384, opened 2015-07-06
Add namespace and make events past tense

### LINKED ISSUE #1383: Placard documentation of events
Currently the placard documentation for events differs from the code. 
screen shot 2015-07-02 at 9 30 38 pm

Documented => coded (line number)
show.fu.placard => shown.fu.placard (230)
hide.fu.placard => hidden.fu.placard (168)
accept.fu.placard => accept (54, 86)
cancel.fu.placard => cancel (56, 89)

interactivellama:
This change still needs to be discussed. Fixes #1383.

This is only a breaking change if developers have used `$().on('accept')` as 'documented' in the source and not `$().on('accept.fu.placard')` and `$().on('cancel.fu.placard')` as the documentation recommends.

Also, these appear to be events that did not get switched to past tense.

**UPDATED**: This is a breaking change for event names of placard.


swilliamset:
fix with past tense.

interactivellama:
We can support and fire **both** since adding an 'ed' will not fire the event twice like just adding a namespace does.

interactivellama:
Updated PR with **DEFINITELY** breaking changes. The callbacks, `onAccept` and `onCancel`, remain intact.

Events are now `accepted.fu.placard` and `cancelled.fu.placard`

index.js
```diff
@@ -440,6 +440,13 @@ define(function (require) {
 		});
 	});
 
+	$('#myPlacard3').on('accept.fu.placard', function() {
```
kevinparkerson: these need to be `accepted.fu.placard` and `cancelled.fu.placard` later on down

interactivellama: Updated. Thanks!

js/placard.js
```diff
@@ -63,30 +63,31 @@
 		constructor: Placard,
 
 		complete: function (action) {
-			var func = this.options['on' + action[0].toUpperCase() + action.substring(1)];
+			var eventName = action === 'accepted' ? 'accept' : 'cancel';
```
swilliamset:
```
var EVENT_MAP = { 'accepted': 'onAccept', 'cancelled': 'onCancel' }; // move up in scope outside function
var func = this.option[EVENT_MAP[action]];
```

swilliamset:
Make the suggested adjustment above otherwise the logic for determining a function name is overly complicated.

js/placard.js
interactivellama: Sounds good. It allows `'maybe':'imThinkingAboutIt'` to be added in the future. j/k

swilliamset:
pushed back to make constants mapper a private variable

interactivellama:
Updated.

PR #1386, opened 2015-07-07
Repeater header re-render fix

### LINKED ISSUE #1385: Repeater does not re-render headers if headers are not present
If using the `clear` method (useful for infinite scrolling), the repeater will only check that the columns have not changed and will not re-render the headers if they are not present.

kevinparkerson:
Fixes #1385  Makes sure headers re-render by checking to see if the thead is there as part of determining whether headers get rendered again.

Also fixed a small bug that was causing setViewOptions to not correctly find the appropriate view options within the repeater core if not named in a specific manner.


kevinparkerson:
not sure what happened here... way more changes than what I made... will re-issue

PR #1404, opened 2015-07-15
Bower install if $INSTALL_BOWER is true

interactivellama:
Adds bash script to handle post-install and installs bower based on environmental variables.


interactivellama:
@swilliamset This is broken http://fuelux-dev.herokuapp.com/ until this PR gets merged.

PR #1405, opened 2015-07-17
Allowing frozen columns to sort

BenjaminNeilDavis:
Frozen columns now trigger the sort of the repeater.
- Had to be more specific on the selectors so it would not look also in the frozen column table for sorted properties because the frozen table does not have data attributes like index attached to it

### DIFF
--- a/index.js
+++ b/index.js
@@ -635,7 +635,7 @@ define(function (require) {
 					list_columnSizing:false,
 					list_columnSyncing: false,
 					list_selectable: false, // (single | multi)
-					list_frozenColumns: 1
+					list_frozenColumns: 2
 				}
 			}
 		});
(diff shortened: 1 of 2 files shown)


interactivellama:
Please add ability to sort to `index.html` page via `index.js`:

```
	var _ = require('underscore');
```

```
var sort = function( data, sortProperty, sortDirection ) {
				var sortedData = _.sortBy( data, function( item ) {
					return item[ sortProperty ];
				} );

				// sort direction
				if ( sortDirection === 'desc' ) {
					sortedData = sortedData.reverse();
				}

				return sortedData;
			};

			resp.items = sort( resp.items, options.sortProperty, options.sortDirection );
```

Probably to both repeaters.

BenjaminNeilDavis:
@interactivellama Has been updating with sorting!

PR #1407, opened 2015-07-20
Add checkbox support for multi select

### LINKED ISSUE #1311: native support of checkbox in multi select repeater

BenjaminNeilDavis:
Added checkbox support for multi select

Fixes #1311, #1421

### DIFF
--- a/index.js
+++ b/index.js
@@ -542,12 +542,14 @@ define(function (require) {
 				{
 					label: 'Common Name',
 					property: 'commonName',
-					sortable: true
+					sortable: true,
+					width: 600
 				},
 				{
 					label: 'Latin Name',
 					property: 'latinName',
-					sortable: true
+					sortable: true,
+					width: 600
 				},
 				{
 					label: 'Appearance',
@@ -630,9 +632,7 @@ define(function (require) {
 					dataSource: function (options, callback) {
 						list(options, callback);
 					},
-					list_selectable: 'multi',
-					list_columnSizing:false,
-					list_columnSyncing: false
+					list_selectable: 'multi'
 				},
 				'thumbnail': {
 					dataSource: function (options, callback) {
@@ -646,8 +646,6 @@ define(function (require) {
 					dataSource: function (options, callback) {
 						list(options, callback);
 					},
-					list_columnSizing:false,
-					list_columnSyncing: false,
 					list_selectable: false, // (single | multi)
 					list_frozenColumns: 1
 				}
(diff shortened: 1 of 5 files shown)


interactivellama:
Please add example using checkbox to `index.html` in order to test.

BenjaminNeilDavis:
@interactivellama the first example will work. All that is needed is multi-select enabled.

interactivellama:
Do we want any kind of set spacing on these checkboxes?

index.js
```diff
@@ -646,8 +648,8 @@ define(function (require) {
 					dataSource: function (options, callback) {
 						list(options, callback);
 					},
-					list_columnSizing:false,
-					list_columnSyncing: false,
+/*					list_columnSizing:false,
```
interactivellama: Please remove comments if not needed.

js/repeater-list.js
```diff
@@ -452,6 +492,19 @@
 		}
 		$row.append($col);
 
+		if (this.viewOptions.list_selectable === 'multi' && columns[columnIndex].property === '@_CHECKBOX_@') {
+			var checkBoxMarkup = '<label class="checkbox-custom checkbox-inline body-checkbox" data-initialize="checkbox">' +
+				'<input class="sr-only" type="checkbox"></label>';
```
interactivellama: Although I'm not aware of any side affects, we may not want to not promote the use of `data-initialize="checkbox"` and manually use `$().checkbox();`. I do know that we ran into issues with auto initializing in scheduler. This will also allow us to determine when the checkbox initializes instead of waiting for a mouseover.

js/repeater-list.js
```diff
@@ -47,7 +47,7 @@
 
 		$.fn.repeater.Constructor.prototype.list_getSelectedItems = function () {
 			var selected = [];
-			this.$canvas.find('.repeater-list table tbody tr.selected').each(function () {
+			this.$canvas.find('.repeater-list table tbody tr.selected, .repeater-list table tbody tr.multi-selected').each(function () {
```
interactivellama: We may want to considering just using `selected` instead `multi-selected` in order to simplify now that we have a wrapper class that will cascade.

interactivellama:
@BenjaminNeilDavis I think this is older code I'm referring to, but is there a reason that the action.html key is a function and not a string? Line 244: `var html = action.html();` is a little confusing because it's so close to `http://api.jquery.com/html/`.

interactivellama:
Please merge in glyphicons into actions dropdown. #1410

interactivellama:
Are the double border line to the left side of actions column intentional? This is with column sizing on.

screen shot 2015-07-27 at 10 45 19 am

BenjaminNeilDavis:
@interactivellama Did you do anything different to get that double line? What page are you looking at? What browser are you using? I am looking at index page in chrome and do not see that

interactivellama:
Reviewed: Assigned to @kevinparkerson for additional review.

interactivellama:
"Auto flip" of the dropdown to cause the actions dropdown to flip up has been added in this PR.

It does not use `data-target` due to wrapping div not having a unique ID, etc. and therefore transverses the all parent DOM notes for `overflow:hidden`.

interactivellama:
@BenjaminNeilDavis It's Chrome. I think it has something to do with column sizing being on.

kevinparkerson:
I see nothing wrong with the code after a look at it. Would like to test it locally, but the clone download percentage is still only at 34% after 20 minutes. Slow today, for some reason. Might check it out later after it finishes.

swilliamset:
@interactivellama and @kevinparkerson approved offline

PR #1410, opened 2015-07-20
Add glyphicons to example page

interactivellama:
Fuel UX does not support MC theme icons. 

Also, `<span>` is typically used within `<a>` and with icons in Bootstrap and Fuel UX.


BenjaminNeilDavis:
Looks good! Thanks for doing that for me

PR #1426, opened 2015-07-29
Added a check to see if there were no items. 

### LINKED ISSUE #1424: Horizontal sizing issue when multi-select is enabled and zero items
Set count in datasource "return" to 0 to duplicate.

screen shot 2015-07-28 at 3 25 22 pm

BenjaminNeilDavis:
If there are no items it will not insert the extra multi select row and will not run multi select functions attached to it.
Fixes #1424


swilliamset:
For future reference move the empty condition first https://github.com/ExactTarget/fuelux/pull/1426/files#diff-707dd2374ec231a926be8e6a6d3c9faaR485

PR #1435, opened 2015-07-31
1434 - repeater list should not add css class sorted to TD of empty row

rbmoore:


rbmoore:
this is a fix for 1434

Issue #1436, opened 2015-07-31
Repeater list - display error message


rbmoore:
It would be nice to be able to display an error message in the repeater list if an error occurs.  The object passed to the datasource callback function would have a new "error" property added to it if an error occurred.  The developer that is using the repeater list control would set obj.error to an error message they deem appropriate.  The Repeater List renderTbody function would then render the error message if it detects one.

![screen shot 2015-07-31 at 5 46 41 pm](https://cloud.githubusercontent.com/assets/4680359/9018441/f609f940-37ac-11e5-95bb-f62499ba1f4f.png)

interactivellama:
This is definitely needed. The issue has been kicked around since December. See #945 (Null/empty state for repeater?). We may want to support more uses than errors alone and create a more generic "custom message" option to support something like FTUX.

rbmoore:
The advantage of having the error property is that we can have specific rendering for errors.  In this case the text is highlighted red.

swilliamset:
@rbmoore merge accepted. additional unit tests would be nice.

PR #1437, opened 2015-07-31
1436

rbmoore:
This is for issue #1436

The code results are depicted in the screen shot tied to the issue.


rbmoore:
If the error idea is accepted,  I can add a commit to this pull request to add unit tests for this new feature.

PR #1445, opened 2015-08-05
correct package version on pushing tag; correct cdn push script

swilliamset:


interactivellama:
I believe these are the lines that will make it make more sense: 
https://github.com/swilliamset/fuelux/blob/grunt-release-updates/Gruntfile.js#L849
https://github.com/ExactTarget/fuelux/blob/master/Gruntfile.js#L866

`updateRelease` is a task that is used to update the variable packageVersion to the new version you are releasing. `pkg.version` is parsed at the initiation/compile of `gruntFile.js` and will be set to the "old" previous version, since you started the release task with the old version listed in `package.json`, not the one you are releasing. 

I guess there's nothing that says we could just overwrite `pkg.version` instead.

PR #1449, opened 2015-08-10
Adding check before doing custom column render. 

### LINKED ISSUE #1447: Repeater custom row render removes checkboxes
image

With list_rowRendered & list_columnRendered

BenjaminNeilDavis:
Making sure the column is not either checkbox or actions column.

Fixes #1447

### DIFF
--- a/js/repeater-list.js
+++ b/js/repeater-list.js
@@ -538,7 +538,7 @@
 			$col.html(checkBoxMarkup);
 		}
 
-		if (this.viewOptions.list_columnRendered && !(columns[columnIndex].property === '@_CHECKBOX_@' || columns[columnIndex].property === '@_ACTIONS_@')) {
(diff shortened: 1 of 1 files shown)


js/repeater-list.js
```diff
@@ -538,7 +538,7 @@
 			$col.html(checkBoxMarkup);
 		}
 
-		if (this.viewOptions.list_columnRendered) {
+		if (this.viewOptions.list_columnRendered && !(columns[columnIndex].property === '@_CHECKBOX_@' || columns[columnIndex].property === '@_ACTIONS_@')) {
```
swilliamset: switch the order. if the column is checkbox or action it doesn't matter if there is a custom render.

BenjaminNeilDavis:
@swilliamset I switched the order

PR #1456, opened 2015-08-11
Make TravisCI use docker containers

interactivellama:
See http://docs.travis-ci.com/user/migrating-from-legacy/


interactivellama:
Oh, wow. That was beautiful. Completed in 40 seconds.

PR #1457, opened 2015-08-11
Make package version dynamic in grunt

interactivellama:
Previous version set at "compile time" and updated once. Now the version is always "fresh." @swilliamset 

Replaces #1445.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -3,11 +3,14 @@
 module.exports = function (grunt) {
 	'use strict';
 
+	function getPackageVersion() {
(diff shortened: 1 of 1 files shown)


swilliamset:
Getting fresh package version, good. Always getting fresh package version even though you haven't changed the version especially in the score of a function, bad. Little overkill.

interactivellama:
Yes, if performance was a concern, this would not be the answer. 

The other option is update it at a certain point in the process, but that mean you are managing state which can easily go bad. _Is this tasking running before or after the update?_

This makes the version stateless and always reads it from the file.

PR #1468, opened 2015-08-17
makes timezone return all $.data() instead of just select values

### LINKED ISSUE #1466: Return time zone ID with name and offset
http://getfuelux.com/javascript.html#scheduler-usage-methods-value

Need third property to return of `id` (or something) which would represent the id of the value the user  chose. For instance, if you called `$('#myScheduler').scheduler('value');` after the user chose "(GMT) Casablanca *" it would return the following timezone object:

```
timeZone: {
    offset: '+02:00',
    id: '2'
    name: "(GMT) Casablanca *"
}
```

Our APIs require us to pass them the `id` rather than the offset or the name.

cormacmccarthy:
fixes #1466

### DIFF
--- a/js/scheduler.js
+++ b/js/scheduler.js
@@ -17,7 +17,7 @@
 		define(['jquery', 'fuelux/combobox', 'fuelux/datepicker', 'fuelux/radio', 'fuelux/selectlist', 'fuelux/spinbox'], factory);
 	} else if (typeof exports === 'object') {
 		// Node/CommonJS
-		module.exports = factory(require('jquery'), require('./combobox'), require('./datepicker'), 
+		module.exports = factory(require('jquery'), require('./combobox'), require('./datepicker'),
 			require('./radio'), require('./selectlist'), require('./spinbox') );
 	} else {
 		// OR use browser globals if AMD is not present
@@ -365,10 +365,7 @@
 
 			var data = {
 				startDateTime: startDateTime,
-				timeZone: {
-					name: timeZone.name,
-					offset: timeZone.offset
-				},
+				timeZone: timezone,
 				recurrencePattern: pattern
 			};
 
--- a/js/scheduler.js
+++ b/js/scheduler.js
@@ -365,7 +365,7 @@
 
 			var data = {
 				startDateTime: startDateTime,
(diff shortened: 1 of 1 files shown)


js/scheduler.js
```diff
@@ -365,10 +365,7 @@
 
 			var data = {
 				startDateTime: startDateTime,
-				timeZone: {
-					name: timeZone.name,
-					offset: timeZone.offset
-				},
+				timeZone: timezone,
```
swilliamset: Looking through this file I don't see where `timezone` (all lower case is being set). Please confirm.

interactivellama:
This should work. Not sure why TravisCI is failing. Scope probably, says `timeZone` isn't defined. 

`var timeZone = this.$timeZone.selectlist('selectedItem');` should return all the data via the [selectlist API](http://getfuelux.com/javascript.html#selectlist-usage-methods-selectedItem) as you say.

Don't forget about `setValue`.

js/scheduler.js
interactivellama: Ah, that's it. Just a typo, should be `timeZone`.

cormacmccarthy:
This should not affect setValue. We aren't actually changing the API, all we are doing is allowing the dev access to data items they set on the value the user selected. This lets them do something like (contrived example) `data-color` on each item to assign colors to timezones and then get the color associated with the timezone the user selected.

interactivellama:
If I want to set the timezone programatically, wouldn't we want to allow the app to set the id, too?

I think the app ignores the actual selectlist options and just selects whatever you feed it via `setValue()`

cormacmccarthy:
Our official example doesn't list or support timeZoneIds. The setValue code is:

image

Trying to add timeZoneIds to the API is well beyond the scope of the change I'm requesting. Even allowing timezone selection by _any other_ data attribute on that field is beyond the scope of my request, although I wouldn't necessarily argue against it. It's just out of scope for what I need _right now_.

swilliamset:
We should refactor to make the `setValue` more flexible
https://github.com/ExactTarget/fuelux/blob/master/js/scheduler.js#L466-L470

PR #1472, opened 2015-08-18
Allow any timzone object keys for scheduler setValue

### LINKED ISSUE #1471: Allow scheduler to accept any number of time zone object keys in SetValue
Refactor scheduler `setValue()` to make less fragile and accept any object key matching and query the timezone selectlist by all the object keys via data-attributes.

See #1468.

interactivellama:
Create jQuery selection string for timezone object based on data-attributes and pass to time zone selectlist. Would love to test if it is an object first, but this is JavaScript.

Fixes #1471.

### DIFF
--- a/js/scheduler.js
+++ b/js/scheduler.js
@@ -458,19 +458,17 @@
 				startDate = currentDate.getFullYear() + '-' + currentDate.getMonth() + '-' + currentDate.getDate();
 			}
 
-			item = 'li[data';
+			// create jQuery selection string for timezone object
+			// based on data-attributes and pass to selectlist
+			item = 'li';
 			if (options.timeZone) {
 				if (typeof (options.timeZone) === 'string') {
-					item += '-name="' + options.timeZone;
+					item += '[data-name="' + options.timeZone + '"]';
 				} else {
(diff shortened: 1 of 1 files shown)


js/scheduler.js
```diff
@@ -458,19 +458,17 @@
 				startDate = currentDate.getFullYear() + '-' + currentDate.getMonth() + '-' + currentDate.getDate();
 			}
 
-			item = 'li[data';
+			// create jQuery selection string for timezone object
+			// based on data-attributes and pass to selectlist
+			item = 'li';
 			if (options.timeZone) {
 				if (typeof (options.timeZone) === 'string') {
 					item += '-name="' + options.timeZone;
```
cormacmccarthy: I think this needs adjusted as well.
```item += '[data-name="' + options.timeZone + '"]';```

cormacmccarthy:
Correct that one thing, and I can verify that this works in AS :)

interactivellama:
Updated. https://github.com/ExactTarget/fuelux/pull/1472/files#diff-8bc171525b04a15b620ce751fc57a0a8R466

PR #1479, opened 2015-08-21
Allow spinner step interval option to be set via data attribute

cormacmccarthy:
Allows for spinner step to be set inside of scheduler where you have no ability to directly influence the spinbox initialization settings.


cormacmccarthy:
The other benefit here is that this keeps having different spinners per recurrence option simple.

https://gist.github.com/cormacmccarthy/45c6af999bf79fd65e8e

PR #1485, opened 2015-08-24
Remove border-radius from wizard first step bottom left

interactivellama:
Remove border-radius from wizard first step bottom left since it is not rounded.


swilliamset:
@chriscorwin please insure testing in snippets and pages examples

PR #1492, opened 2015-08-27
Allow default unit to be added to spinbox if not present

### LINKED ISSUE #1490: Allow units to added after parsing
Allow user to not type units and allow a default unit to be specified in the options (such as `%` or `px`) to be added if there are no units.

interactivellama:
I'd call this a bug, since I think spinbox was meant to work this way, but it is a new initialization option.

Fixes #1490.

### DIFF
--- a/js/spinbox.js
+++ b/js/spinbox.js
@@ -131,7 +131,7 @@
 			// if set and default unit if not already present, 
 			// and is an allowed unit, then add default unit
 			if (this.options.defaultUnit !== '' && 
-					this.options.defaultUnit !== value.slice(-2) &&
+					this.options.defaultUnit !== value.slice(-Math.abs(this.options.defaultUnit.length)) &&
 					this.isUnitLegal(this.options.defaultUnit)) {
 				value = value + this.options.defaultUnit;
 			}
--- a/js/spinbox.js
+++ b/js/spinbox.js
@@ -128,6 +128,10 @@
 
 		output: function (value, updateField) {
 			value = (value + '').split('.').join(this.options.decimalMark);
+			// if set and default unit if not already present, add default unit
+			if (this.options.defaultUnit !== '' && this.options.defaultUnit !== value.slice(-2)) {
+				value = value + this.options.defaultUnit;
+			}
 			updateField = (updateField || true);
 			if (updateField) {
 				this.$input.val(value);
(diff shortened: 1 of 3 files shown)


swilliamset:
White space trouble. Input starts with a space between the number and the unit but them removes the space.
sep 04 2015 15 50

Also if I set the default unit to something besides the accepted units including `null` I get weird behavior. 
```
	$('#mySpinbox2').spinbox({
		value: '10 px',
		min: 0,
		max: 10,
		step: 0.1,
		// decimalMark: ',',
		units: ['px'],
		defaultUnit: 'ouch'
	});
```
sep 04 2015 15 54
```
	$('#mySpinbox2').spinbox({
		value: '10 px',
		min: 0,
		max: 10,
		step: 0.1,
		// decimalMark: ',',
		units: ['px'],
		defaultUnit: null
	});
```
sep 04 2015 15 55

Granted these are fringe cases but should have better handling.

Feel free to use the config blocks above to write additional unit tests.

interactivellama:
The defaults in our apps right now are no space whether tradition or intentional.
screen shot 2015-09-08 at 4 15 25 pm

interactivellama:
Updated with non-allowed default unit test.

I am not adding null. Non-strings should not be used, nor accounted for.

For instance `options.units` better be an array, since it used `.length` without checking to see if it's an array.

cormacmccarthy:
Breaks for any 3+ digit units

	$('#mySpinbox2').spinbox({
		value: '1,0pxs',
		min: 0,
		max: 10,
		step: 0.1,
		decimalMark: ',',
		units: ['pxs'],
		defaultUnit: 'pxs'
	});

image

interactivellama:
Updated. Thanks.

js/spinbox.js
```diff
@@ -128,6 +128,13 @@
 
 		output: function (value, updateField) {
 			value = (value + '').split('.').join(this.options.decimalMark);
+			// if set and default unit if not already present, 
+			// and is an allowed unit, then add default unit
+			if (this.options.defaultUnit !== '' && 
+					this.options.defaultUnit !== value.slice(-2) &&
```
cormacmccarthy: You'll probably want to adjust this to account for units longer than 2

interactivellama: Refresh the page.

cormacmccarthy:
Please add a unit test for some unit/default unit combo that is 3+ digits and increments the spinbox at least once.

interactivellama:
@cmcculloh Added test that uses `rem` and increments. Great suggestions!

PR #1493, opened 2015-08-28
Fusion site requires mc theme styled documentation

dwaltz:


swilliamset:
Needs to be wrapped in a condition to not theme only when needed (not getfuelux.com).

swilliamset:
Correction this layout is specific to internal documentation and not getfuelux.com so good to go

PR #1504, opened 2015-09-04
Font family change for fuelux-fusion site

dwaltz:


interactivellama:
Should this be SF sans?

dwaltz:
Hey @interactivellama thanks for bringing that up. Due to your comment design has re-assessed font choices and we will not be using SalesforceSans. The above pull request includes these changes now.

PR #1506, opened 2015-09-08
Superpicker

cormacmccarthy:
Adds a new Fuel UX control, "superpicker", which is not unlike a placard/modal had a baby ("Contextual Modal" was a suggested name at one point). It allows for being triggered from elements _other than_ just an input, as well as placing any arbitrary controls inside of it.


dist/js/fuelux.js
```diff
@@ -2956,6 +2959,271 @@
...
+
+			this.isInput = this.$field.is( 'input' );
+
+			this.$field.on( 'focus.fu.superpicker', $.proxy( this.show, this ) );
+			this.$field.on( 'keydown.fu.superpicker', $.proxy( this.keyComplete, this ) );
+			this.$trigger.on( 'focus.fu.superpicker', $.proxy( this.show, this ) );
+			this.$accept.on( 'click.fu.superpicker', $.proxy( this.complete, this, 'accepted' ) );
+			this.$cancel.on( 'click.fu.superpicker', function( e ) {
+				e.preventDefault();
+				self.complete( 'cancelled' );
```
tweettypography: Super minor, but do we typically spell with two l's?

dist/js/fuelux.js
```diff
@@ -2956,6 +2959,271 @@
...
+				this.$popup.css( 'width', ( this.options.width ) ? this.options.height : DEFAULT_WIDTH + 'px' );
+
+				this.$element.trigger( 'shown.fu.superpicker', this.actualValue );
+				if ( this.actualValue !== null ) {
+					this.actualValue = null;
+				}
+
+				this.$field.attr( 'disabled', 'disabled' );
+
+				this.clickStamp = new Date().getTime() + ( Math.floor( Math.random() * 100 ) + 1 );
```
tweettypography: Why generate this each time on show instead of in the constructor? (And for the matter, I assume the purpose of the `clickStamp` is to isolate your `.off` calls, but maybe you could achieve the same effect by including the function in that call?)

dist/js/fuelux.js
```diff
@@ -2956,6 +2959,271 @@
...
+
+			return ( methodReturn === undefined ) ? $set : methodReturn;
+		};
+
+		$.fn.superpicker.defaults = {
+			onAccept: undefined,
+			onCancel: undefined,
+			externalClickAction: 'cancelled',
+			externalClickExceptions: [],
+			explicit: false,
```
tweettypography: After reading the code I believe this means you have to explicitly close the pop-up instead of just clicking elsewhere, but I wonder if it could be more clearly named?

dist/js/fuelux.js
```diff
@@ -2956,6 +2959,271 @@
...
+			revertOnCancel: -1 //negative 1 will check for an '.placard-accept' button. Also can be set to true or false
```
tweettypography: Is this -1 option really necessary?

dist/js/fuelux.js
```diff
@@ -2956,6 +2959,271 @@
...
+				this.$element.addClass( 'showing' );
+
+				this.$popup.css( 'top', ( this.$field.outerHeight( true ) + 4 ) + 'px' );
+
+				this.$popup.css( 'height', ( this.options.height ) ? this.options.height : DEFAULT_HEIGHT + 'px' );
+				this.$body.css( 'height', ( ( this.options.height ) ? this.options.height : DEFAULT_HEIGHT ) - 73 + 'px' );
+				this.$popup.css( 'width', ( this.options.width ) ? this.options.height : DEFAULT_WIDTH + 'px' );
+
+				this.$element.trigger( 'shown.fu.superpicker', this.actualValue );
+				if ( this.actualValue !== null ) {
```
tweettypography: What does this do?

dist/js/fuelux.js
```diff
@@ -2956,6 +2959,271 @@
...
+					return;
+				}
+
+				other = $( document ).find( '.superpicker.showing' );
+				if ( other.length > 0 ) {
+					if ( other.data( 'fu.superpicker' ) && other.data( 'fu.superpicker' ).options.explicit ) {
+						return;
+					}
+
+					other.superpicker( 'externalClickListener', {}, true );
```
tweettypography: Just curious, why a second argument instead of an object property?

dist/js/fuelux.js
```diff
@@ -2956,6 +2959,271 @@
...
+			} );
+
+			return ( methodReturn === undefined ) ? $set : methodReturn;
+		};
+
+		$.fn.superpicker.defaults = {
+			onAccept: undefined,
+			onCancel: undefined,
+			externalClickAction: 'cancelled',
+			externalClickExceptions: [],
```
tweettypography: What's the use case for these exceptions?

dist/js/fuelux.js
cormacmccarthy: This is consistent with Placard.

dist/js/fuelux.js
cormacmccarthy: Agreed. This is consistent with Placard.

PR #1510, opened 2015-09-10
If only one selected return its object instead of array

### LINKED ISSUE #1509: repeater action doesn't distinguish between bulk action and inline action
https://github.com/ExactTarget/fuelux/blob/master/js/repeater-list.js#L331

BenjaminNeilDavis:
Fixes #1509

### DIFF
--- a/js/repeater-list.js
+++ b/js/repeater-list.js
@@ -335,6 +335,9 @@
 					rowData: clickedRow.data('item_data')
 				});
 			}
+			if (selectedObj.length === 1) {
+				selectedObj = selectedObj[0];
+			}
 
 			if (actionObj.clickAction) {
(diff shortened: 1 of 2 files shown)


js/repeater-list.js
```diff
@@ -335,6 +335,9 @@
 					rowData: clickedRow.data('item_data')
 				});
 			}
+			if (selectedObj.length === 1) {
+				selectedObj = selectedObj[0];
+			}
```
swilliamset: row level actions at line 306 above passes a single object to `list_getActionItems`.

change the condition to check if `selected` is an array or an obj. `if (!<obj>)` do for loop. `else selectedObj = selected`

PR #1536, opened 2015-09-22
makes step increment work properly in schedulerfuelux

cormacmccarthy:
Depends on #1531

### DIFF
--- a/index.html
+++ b/index.html
@@ -2445,7 +2445,7 @@
 							</button>
 						</div>
 					</div>
-					<div class="spinbox" id="mySpinbox3" data-step="5">
+					<div class="spinbox" id="mySpinbox3">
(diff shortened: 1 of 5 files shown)


futuremint:
I can type in a number that is less than the limit, and it will get rounded down _below_ the "one unit" threshold to 0.  Basically using the "convenience rounding to nearest increment" logic to defeat the "don't go below a single increment unit" logic :).  Its like UI control judo :kimono:

cormacmccarthy:
Congratulations @futuremint you have found a spinbox bug! That's unrelated to scheduler and unrelated to this pull request, but, I'm fixing it now anyways :)

futuremint:
:joy_cat: do I get a cookie now?

Issue #1463, opened 2015-08-17
[SelectList] Width does not consider images within list elements


scott-joe:
I was having PhantomJS issues when running the tests to validate that this would work as intended. But I've got a select list with icons in the `a` tags (used in prod, but may not be a real pattern). For whatever reason, the icons are not correctly interpreted as part of the width.

<img width="83" alt="screen shot 2015-08-16 at 10 16 24 pm" src="https://cloud.githubusercontent.com/assets/2026866/9296915/bc693998-4464-11e5-96ac-de30070d0aa8.png">

https://jsbin.com/ginoxukuwe/1/edit?html,output

I thought perhaps changing the `a` to `li` to perhaps capture the entire width, but that may not even be the solution.

interactivellama:
@scott-joe Will look into. I can tell you the resize is created by cloning the selectlist and then calling outerWidth() on each item and choosing the largest width. 

How do images work in "jQuery memory" with elements that are never added to the DOM? I would imagine that the `img src` is never called/loaded until added to the DOM and this may be causing the issue.

Looks like this is the culprit! LN158 `sizer.find('.selected-label').text($(this).text());`. That's going to strip the non-text out. Should probably be `.html()`.

scott-joe:
Awesome. Glad it wasn't just in my head. Thanks for looking into it! And I learned something about "jQuery memory" in the process.

PR #1537, opened 2015-09-23
Fix width when using images within selectlist

scott-joe:
My terminal is borked so you'll have to check the solution, but I just did what you suggested in the issue thread #1463

### DIFF
--- a/js/selectlist.js
+++ b/js/selectlist.js
@@ -59,7 +59,7 @@
 				var $this = $(this);
 				// attach key listener when dropdown is shown
 				$(document).on('keypress.fu.selectlist', function(e){
-					
+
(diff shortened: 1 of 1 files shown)


futuremint:
Please ensure that this fix is _actual_ and not _probable_.  Once you're sure this _is the fix_, we'll review it.

swilliamset:
hold until full validation as suggested by woodward above

cormacmccarthy:
image

Selectlist resizes vertically as well as horizontally. @jamin-hall is this desired? Or should it be vertically static?

cormacmccarthy:
Needs some vertical alignment love.

PR #1538, opened 2015-09-23
Fixes disabling of selectable / multi select / actions

### LINKED ISSUE #1534: selectable repeater does not respect disabled state
use jsbin for testing. currently fuelux.min.js points to local. http://jsbin.com/kucifivufo/1/edit?html,js,output

BenjaminNeilDavis:
Fixes #1534

### DIFF
--- a/js/repeater-list.js
+++ b/js/repeater-list.js
@@ -594,6 +594,7 @@
 		if (sortable) {
 			$both.addClass('sortable');
 			$div.on('click.fu.repeaterList', function () {
+				if(self.isDisabled) { return; }
 				self.list_sortProperty = (typeof sortable === 'string') ? sortable : columns[index].property;
 				if ($div.hasClass('sorted')) {
 					if ($span.hasClass(chevUp)) {
@@ -655,6 +656,7 @@
 			$row.data('item_data', rows[index]);
 
 			$row.on('click.fu.repeaterList', function () {
+				if(self.isDisabled) { return; }
 				var $item = $(this);
 				var index = $(this).index();
 				index = index + 1;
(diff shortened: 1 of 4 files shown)


js/repeater-list.js
```diff
@@ -594,6 +594,7 @@
 		if (sortable) {
 			$both.addClass('sortable');
 			$div.on('click.fu.repeaterList', function () {
+				if(self.isDisabled) return;
```
kevinparkerson: this is technically correct, but as per our style guide we are supposed to write it as follows:
`if (self.isDisabled) { return; }`

js/repeater-list.js
```diff
@@ -655,6 +656,7 @@
 			$row.data('item_data', rows[index]);
 
 			$row.on('click.fu.repeaterList', function () {
+				if(self.isDisabled) return;
```
kevinparkerson: this is technically correct, but as per our style guide we are supposed to write it as follows:
if (self.isDisabled) { return; }

kevinparkerson:
Looks good outside minor style guide compliance changes! :+1:

futuremint:
@BenjaminNeilDavis looks like there is an issue with the unit tests: https://travis-ci.org/ExactTarget/fuelux/builds/81997536#L519 Can you please fix?

cormacmccarthy:
https://github.com/cormacmccarthy/fuelux/tree/BenjaminNeilDavis-issue-1534-diable-repeater

I've updated the syntax of your async tests @BenjaminNeilDavis. However, there is still some sort of race condition keeping your tests from completing consistently and therefore when run in the context of all tests it fails (however, when run alone, it succeeds).

You can check the tests locally by running `grunt servefast` and then go http://localhost:8000/test/

PR #1556, opened 2015-10-05
keep end date from being before start date when start date changed

### LINKED ISSUE #1487: Keep end date from being before start date
1. Go to http://getfuelux.com/javascript.html#scheduler-examples-live-demo
2. Select a future start date
3. Select a repeat (e.g. hourly)
4. Select end on date

Expected results:
End on date sets itself to just after the first possible run (so, weekly repeat, the end on date would set itself to start date + 1 week. Hourly repeat, end on date would set itself to start date + 1 hour)

Actual results:
End on date sets itself to current date (which is prior to start date)

image

Any time the user changes the start date, if it is greater than the end date, we need to adjust the end date by the repeat increment.

cormacmccarthy:
fixes #1487

### DIFF
--- a/js/scheduler.js
+++ b/js/scheduler.js
@@ -68,6 +68,7 @@
 		//initialize sub-controls
 		this.$element.find('.selectlist').selectlist();
 		this.$startDate.datepicker(this.options.startDateOptions);
+		this.$startDate.on('changed.fu.datepicker', $.proxy(this.startDateChanged, this));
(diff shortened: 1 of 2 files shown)


interactivellama:
JShint error.

https://travis-ci.org/ExactTarget/fuelux/builds/83795399#L412

swilliamset:
screen shot 2015-10-06 at 10 53 20 am

PR #1558, opened 2015-10-06
.jshintrc vs jshint options, and why you should be using the former

cormacmccarthy:


grunt/config/jshint.js
```diff
@@ -1,33 +1,7 @@
 module.exports = {
 	options: {
-		boss: true,
-		browser: true,
-		curly: false,
-		eqeqeq: true,
-		eqnull: true,
-		globals: {
-			jQuery: true,
```
futuremint: This option was removed... intentional?

cormacmccarthy: Yes, that is what happened.

PR #1559, opened 2015-10-06
Font updates for fusion

dwaltz:


dwaltz:
Hey @interactivellama can you review these changes really quick? We need to get them in as soon as possible. they should only effect fuelux-fusion-site styles.

PR #1563, opened 2015-10-08
Fixes markup and styling issues with Scheduler

### LINKED ISSUE #1511: Scheduler Markup and Alignment
Adjust the scheduler markup for proper alignment.
screenshot 2015-09-11 10 12 06

- start date and end date fields should be the same width
- end date should be displayed inline and float to the right of the end selectlist (ie they shouldn't be in separate columns)

scott-joe:
Fixes #1511

The markup was all over the place. I reworked the columns to work across -lg -sm and -xs sizes and debugged redundant classes resulting in different rendering widths on two "identical" inputs.

This does not however cover MCTHeme styling. That'll have to be another issue & PR on that repo.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1329,7 +1329,7 @@
 			<section id="scheduler">
 				<h2>Scheduler</h2>
 				<div class="thin-box">
-					<div id="myScheduler" class="form-horizontal container-fluid scheduler" role="form" data-initialize="scheduler">
+					<div id="MyScheduler" class="form-horizontal container-fluid scheduler" role="form" data-initialize="scheduler">
 						<div class="row start-datetime">
 							<label class="col-sm-2 control-label scheduler-label" for="myStartDate">Start Date</label>
 							<div class="col-sm-10">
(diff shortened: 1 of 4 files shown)


cormacmccarthy:
This is a great start! The "official markup" resides in /markup/scheduler.html. Also, gh-pages documentation markup will need changed (since that's what people actually copy from).

scott-joe:
Ah okay. I didn't know that. I'll do that and update the PR in the morning.

On Thursday, October 8, 2015, Christopher McCulloh <notifications@github.com>
wrote:

> This is a great start! The "official markup" resides in
> /markup/scheduler.html. Also, gh-pages documentation markup will need
> changed (since that's what people actually copy from).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ExactTarget/fuelux/pull/1563#issuecomment-146722105>.
>


-- 

*Scott Williams*

*Sr. Front-End Engineer, UX Architecture*UXDS, Salesforce Marketing Cloud

scott-joe:
Files are updated. I'll have to update `gh-pages` separately

PR #1571, opened 2015-10-14
Adds jQuery event to action item callback

### LINKED ISSUE #1570: Repeater list_actions, change callback function in clickAction to be an event object instead
Current callback function is always an empty function, having the event object available in this function would be much more useful. Would be able to call preventDefault(), etc...

cormacmccarthy:
Closes #1570


futuremint:
Please add a unit test that covers this additional functionality.

PR #1572, opened 2015-10-14
removes unnecessary blank element from header

### LINKED ISSUE #1568: Repeater with list_selectable: single and list_actions is not generating the header row properly
http://codepen.io/anon/pen/KdvJrP?editors=101

Using list_selectable: single with list_actions creates a bad element in the header that does not have the proper styling.

There is also an issue with alignment of the action column buttons when a row spans multiple lines.

cormacmccarthy:
closes #1568 

I could discern no reason for having this span present, so, I removed it, which fixed the issue.

### DIFF
--- a/js/repeater-list.js
+++ b/js/repeater-list.js
@@ -270,8 +270,12 @@
 					$actionsColumn.find('thead .btn').attr('disabled', 'disabled');
 				}
 				else {
-					var label = this.viewOptions.list_actions.label || '<span class="actions-hidden">a</span>';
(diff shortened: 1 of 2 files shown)


futuremint:
Checking this in IE9 now...

futuremint:
This fix appears to actually make the problems worse, here is Chrome when scrolling with `list_selectable: 'single'` on the index page.
screen shot 2015-10-15 at 3 33 49 pm

futuremint:
Ok, looks like this PR doesn't fix this error.  I'm going to merge this because it at least gets us some more deterministic behaviors we can fix going forward.

PR #1574, opened 2015-10-16
Gh1569   repeater resize issues

### LINKED ISSUE #1569: Repeater with list_actions does not resize horizontally properly
If you load a repeater in a half-width window, then resize your window to a larger size the columns will not resize to fill the space between the right-aligned list_action buttons. Refer to this screenshot: 
screen shot 2015-10-13 at 11 08 35 am

cormacmccarthy:
fixes #1569 

The only actual functionality changes that I made here were to make the whole thing re-render if there are frozen or action columns on resize.

All of the other changes were refactoring trying to determine how everything actually worked to determine how to even fix the issue. re-rendering ended up being the cleanest quickest way.

### DIFF
--- a/js/repeater.js
+++ b/js/repeater.js
@@ -636,12 +636,15 @@
 					bottom: this.$viewport.css('margin-bottom'),
 					top: this.$viewport.css('margin-top')
 				};
-				height = ((staticHeight === 'true' || staticHeight === true) ? this.$element.height() : parseInt(staticHeight, 10)) -
-				this.$element.find('.repeater-header').outerHeight() -
-				this.$element.find('.repeater-footer').outerHeight() -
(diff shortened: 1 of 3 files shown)


js/repeater.js
```diff
@@ -116,7 +116,7 @@
 			self.resizeTimeout = setTimeout(function () {
 				self.resize();
 				self.$element.trigger('resized.fu.repeater');
-			}, 75);
+			}, 500);//any faster and you get weird catastrophic errors with the header of the list repeater
```
futuremint: So, this makes the UX inferior when resizing, because its slow and flickers when you resize.  I tried adding some code to `resize` to debounce calls to it, but it didn't fix the weird header resizing errors. So... I guess this has to be this way for now :(

PR #1590, opened 2015-10-29
accidentally changed a jquery html to append, caused issues in single…

### LINKED ISSUE #1589: Repeater has horizontal overflow and when scrolled borders are messed up
image

http://codepen.io/anon/pen/KdvJrP?editors=101

Reported by @dkilgore

cormacmccarthy:
… select mode, fixed.

fixes #1589

### DIFF
--- a/js/repeater-list.js
+++ b/js/repeater-list.js
@@ -270,16 +270,16 @@
 				} else {
 					var labelText = this.viewOptions.list_actions.label || '';
 
-					var $labelOverlay = $('<div class="repeater-list-heading empty">' + labelText + '</div>');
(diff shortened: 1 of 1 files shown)


cormacmccarthy:
I think a better solution than what we have now would be to remove the action columns header completely somehow, because there’s still two bugs going on here:

1) When all of the columns fit properly and the repeater doesn't need horizontal scroll, you still can scroll horizontally a little bit:
image

2) When all of the columns don't fit properly, and the empty placeholder is over a column's label, the label gets cut off:
image

However, this fix gets us a better overall experience in the short term.

PR #1591, opened 2015-10-29
Add refreshFolder method to tree

### LINKED ISSUE #1592: Add refreshFolder method to tree
This method will delete folder children from the DOM and reopen the chosen folder, thus calling the dataSource again for it.

interactivellama:
- Add `refreshFolder` method in order to bust "DOM cached" folder children and reload from `dataSource`.
- Unit test makes sure that different nodes are present after refresh by using random numbers on `id`.

Fixes #1592.

*UI pattern update*
- This assumes that `refreshFolder` is a background process, not something initiated by the user-- therefore "Loading..." will not display when folder is refreshed when closed. 
- Developers can use the `refreshedFolder.fu.tree" event to dismiss user feedback if they wish to. 
- A folder will show "Loading..." when folder is currently open for the user, due to the "broken looking" state that would be displayed if "Loading..." was not present when the "folder children disappear" while requesting data from the server.

@jamin-hall Might want to check on this when you return.

Side note: We may want to wait until the data is loaded from the server to actually empty out the folder if "loading..." is not present (this would remove the "broken state" possibility.

### DIFF
--- a/js/tree.js
+++ b/js/tree.js
@@ -87,13 +87,16 @@
 			this.populate(this.$element);
 		},
 
-		populate: function populate($el) {
+		populate: function populate($el, isBackgroundProcess) {
 			var self = this;
 			var $parent = ($el.hasClass('tree')) ? $el : $el.parent();
 			var loader = $parent.find('.tree-loader:eq(0)');
 			var treeData = $parent.data();
+			isBackgroundProcess = isBackgroundProcess || false;	// no user affordance needed (ex.- "loading")
 
-			loader.removeClass('hide hidden'); // hide is deprecated
+			if (isBackgroundProcess === false) {
+				loader.removeClass('hide hidden'); // hide is deprecated
+			}
 			this.options.dataSource(treeData ? treeData : {}, function (items) {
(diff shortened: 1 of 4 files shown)


js/tree.js
```diff
@@ -423,9 +428,20 @@
...
+		},
+
+		// This refreshes the children of a folder. Please destroy and re-initilize control for "root level" refresh.
+		// The data of the refreshed folder is not updated. This control's architecture only allows updating of children.
+		// Also, this function implies that the end-user wants the folder open.
+		refreshFolder: function refreshFolder($el) {
+				var $treebranch = $el.closest('.tree-branch');
+				var $refreshedBranch;
+				$treebranch.find('.tree-branch-children').eq(0).empty();
+				$refreshedBranch = this._discloseFolder($el.find('.tree-branch-header'));
```
swilliamset: Don't need to actually open the folder. Just need to refresh the content in the folder. so just empty and run populate?

interactivellama: We may want to force open because of our "Loading..." pattern.

PR #1598, opened 2015-11-05
added support for search on keypress and fixed issue with clearing ou…

### LINKED ISSUE #1244: Repeater deletes value of search field at the second ENTER keypress
The repeater deletes the value of the search field is ENTER is pressed twice. You get back to an empty search field. You can check this easily on the example page. The first ENTER selects correctly a sublist, the second ENTER deletes the search field.

While this is only slightly puzzling in the example shown on the site, it becomes completely disconcerting if you attach a completer. When the user clicks on an entry of the completer, we trigger the repeater so that it shows the correct result. But if the user hits return again, the query is lost.

We think that hitting ENTER should just trigger the repeater (possibly checking that some element of the filter interface has changed).

mbeard:
…t unchanged search value

removed clearing search on consecutive search (double enter) and unit test

fixes #1244

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -111,14 +111,20 @@
 		},
 
 		keypress: function (e) {
-			if (e.which === 13) {
-				// enter pressed
+			var ENTER_KEY_CODE = 13;
+			var TAB_KEY_CODE = 9;
+			var ESC_KEY_CODE = 27;
+
+			if (e.which === ENTER_KEY_CODE) {
 				e.preventDefault();
 				this.action();
 			}
-			else if(e.which === 9) {
-				// tab pressed
+			else if(e.which === TAB_KEY_CODE) {
 				e.preventDefault();
+			}
+			else if(e.which === ESC_KEY_CODE) {
+				e.preventDefault();
+				this.clear();
 			}
 			else if(this.options.searchOnKeyPress) {
 				// search on other keypress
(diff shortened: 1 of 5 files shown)


js/search.js
```diff
@@ -87,52 +90,39 @@
 
 		action: function () {
 			var val = this.$input.val();
-			var inputEmptyOrUnchanged = (val === '' || val === this.activeSearch);
-
-			if (this.activeSearch && inputEmptyOrUnchanged) {
```
swilliamset: instead of the `Unchanged` part of  `inputEmptyOrUnchanged`

js/search.js
```diff
@@ -87,52 +90,39 @@
...
 		},
 
-		keypressed: function (e) {
-			var remove = 'glyphicon-remove';
-			var search = 'glyphicon-search';
-			var val;
-
+		keypress: function (e) {
 			if (e.which === 13) {
+				// enter pressed
```
swilliamset: Nice to have:
instead of the `Unchanged` part of  `inputEmptyOrUnchanged` add a `escape` event listener for `e.which === 27`

Also add comments above each number or create a constant so code is easier to read
```
var ENTER_KEY_CODE = 13;
var TAB_KEY_CODE = 9;
var ESC_KEY_CODE = 27;
```

PR #1603, opened 2015-11-09
Allow window override in dropdown auto-flip

interactivellama:
This allows explicitly setting auto-flip container to window via `data-target=window`.

### DIFF
--- a/js/dropdown-autoflip.js
+++ b/js/dropdown-autoflip.js
@@ -97,31 +97,33 @@
 	}
 
 	function _getContainer(element) {
-		var containerElement, isWindow;
+		var targetSelector = element.attr('data-target');
+		var isWindow = true;
+		var containerElement;
 
-		// manual override
-		if (element.attr('data-target')) {
-			containerElement = element.attr('data-target');
-			isWindow = false;
-		} else {
-			// default to window otherwise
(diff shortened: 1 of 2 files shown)


swilliamset:
Reconsider with less condition nesting

interactivellama:
Updated with flatter decision tree @futuremint @swilliamset

js/dropdown-autoflip.js
```diff
@@ -97,26 +97,27 @@
...
-			$.each(element.parents(), function (index, value) {
-				if ($(value).css('overflow') !== 'visible') {
-					containerElement = value;
+		// parse decision tree
+		var targetSelector = element.attr('data-target') || false;
+		var nonWindowTargetChosen = targetSelector !== 'window' && targetSelector;
+		var noTargetChosen = targetSelector !== 'window' && !targetSelector;
+
+		if (nonWindowTargetChosen) {
+			targetedElement = $(targetSelector);
```
futuremint: Is this an accidental global?

kevinparkerson: I was just going to comment on this... @interactivellama, when I run `grunt test` jshint catches this and says it is undefined

PR #1606, opened 2015-11-11
support `end === true` attribute to prevent infinite scroll datasource call (fix #1605)

### LINKED ISSUE #1605: Repeater with infinite scrolling does not respect "end:true" property in callback
Going off the documentation here:
http://getfuelux.com/javascript.html#infinite-scroll

Once i retrieve the last page of data the callback is called with dataSource similar to this
``` js
{
    end: true,
    items: ['Array', 'of', 'last', 'items', 'to', 'display', 'in', 'repeater']
}
```

However the infinite repeater continues to make calls for additional data. At this point i have to leave the items array undefined otherwise i get the "No Data" text repeated in the grid. This unfortunately produces an exception inside the repeater code. Not calling the callback will leave the spinner in place as well.

The expected result was once i returned "end:true" I would no longer receive calls into dataSource.

Here is some sample code demonstrating this issue. If you turn on debugging you will see a console error once you try and scroll past the last page of data.

Sample Code:
https://gist.github.com/dreitman-sf/f6fc01828e0b5c266b3f

Sample Code in Action:
http://bl.ocks.org/dreitman-sf/raw/f6fc01828e0b5c266b3f/

swilliamset:
use [this](http://bl.ocks.org/swilliamset/9307df68bd231a414e27) for testing

it points to local build

### DIFF
--- a/js/repeater.js
+++ b/js/repeater.js
@@ -366,7 +366,7 @@
 
 			this.currentPage = (page !== undefined) ? page : NaN;
 
-			if (data.end || (this.currentPage + 1) >= pages) {
+			if (data.end === true || (this.currentPage + 1) >= pages) {
 				this.infiniteScrollingCont.infinitescroll('end', end);
 			}
 		},
--- a/js/repeater.js
+++ b/js/repeater.js
@@ -366,7 +366,7 @@
 
 			this.currentPage = (page !== undefined) ? page : NaN;
 
-			if ((this.currentPage + 1) >= pages) {
+			if (data.end || (this.currentPage + 1) >= pages) {
 				this.infiniteScrollingCont.infinitescroll('end', end);
 			}
 		},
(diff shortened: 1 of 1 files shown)


swilliamset:
in spite of using accurate paging to work around this problem the documentation does say `{end: true}` is supported

swilliamset:
Not sure if this or modifying the call to `infiniteScrollingCallback` (https://github.com/ExactTarget/fuelux/blob/3.12.0/js/repeater.js#L596) is the best approach.

I'd almost prefer `infiniteScrollingCallback` to get the data so the infinite scroll specific logic can be kept with https://github.com/ExactTarget/fuelux/blob/master/js/infinite-scroll.js

js/repeater.js
```diff
@@ -366,7 +366,7 @@
 
 			this.currentPage = (page !== undefined) ? page : NaN;
 
-			if ((this.currentPage + 1) >= pages) {
+			if (data.end || (this.currentPage + 1) >= pages) {
```
kevinparkerson: I like this idea; however, the logic needs to be changed to `(data.end === true)`. That way the current way of handling it would still work, but this just simplifies things :)

PR #1614, opened 2015-11-13
Run `npm update` after wget on TravisCI

interactivellama:
This should help tests run for @greenkeeperio-bot PR's that actually matter and use the versions that were changed.

Currently, we install a cached tarball of Node modules from our Heroku dev app / edge server, but it only gets updated after merges into master.


interactivellama:
Looks like we added about 1m20s to the CI time. Worth it though if we don't have to pull down and GK PR's and test locally.

PR #1625, opened 2015-11-23
Repeater Resize Improvement

kevinparkerson:
Solves issues with repeater resizing by simply resizing columns (including frozen and action) instead of re-rendering. Related to issues 1568 & 1569

### DIFF
--- a/js/repeater-list.js
+++ b/js/repeater-list.js
@@ -79,7 +79,7 @@
 		$.fn.repeater.Constructor.prototype.list_setSelectedItems = function (items, force) {
 			var selectable = this.viewOptions.list_selectable;
 			var self = this;
-			var data, i, $item, l;
+			var data, i, $item, length;
 
 			//this function is necessary because lint yells when a function is in a loop
 			function checkIfItemMatchesValue () {
@@ -113,14 +113,14 @@
 			}
 
 			if (force === true || selectable === 'multi') {
-				l = items.length;
+				length = items.length;
 			} else if (selectable) {
-				l = (items.length > 0) ? 1 : 0;
+				length = (items.length > 0) ? 1 : 0;
 			} else {
-				l = 0;
+				length = 0;
 			}
 
-			for (i = 0; i < l; i++) {
+			for (i = 0; i < length; i++) {
 				if (items[i].index !== undefined) {
 					$item = this.$canvas.find('.repeater-list table tbody tr:nth-child(' + (items[i].index + 1) + ')');
 					if ($item.length > 0) {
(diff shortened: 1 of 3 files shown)


js/repeater.js
```diff
@@ -638,15 +638,12 @@
...
-				var bottomMargin = (viewportMargins.bottom === 'auto') ? 0 : parseInt(viewportMargins.bottom, 10);
-				var topMargin = (viewportMargins.top === 'auto') ? 0 : parseInt(viewportMargins.top, 10);
-				height = staticHeightValue - headerHeight - footerHeight - bottomMargin - topMargin;
-
-				this.$viewport.outerHeight(height); // but WHY are we setting the outerHeight of the viewport to this???
+				height = ((staticHeight === 'true' || staticHeight === true) ? this.$element.height() : parseInt(staticHeight, 10)) -
+					this.$element.find('.repeater-header').outerHeight() -
+					this.$element.find('.repeater-footer').outerHeight() -
+					((viewportMargins.bottom === 'auto') ? 0 : parseInt(viewportMargins.bottom, 10)) -
+					((viewportMargins.top === 'auto') ? 0 : parseInt(viewportMargins.top, 10));
```
futuremint: This giant expression needs to be broken out into clarifying variables.  The rewrite code that you delete was much easier to read and this needs to be unpacked the same way.

js/repeater-list.js
```diff
@@ -881,79 +826,69 @@
...
-			sizeColumns.call(this, $headerRow);
+			sizeColumns.call(this, $tr);
 		}
-	};
+	}
 
-	var sizeColumns = function sizeColumns ($tr) {
-		var autoGauge = [];
+	function sizeColumns ($tr) {
+		var auto = [];
```
futuremint: This variable name needs to be clearer, "auto" is too short and does not describe what it holds and what it is used for.

js/repeater-list.js
```diff
@@ -881,79 +826,69 @@
...
-		if (self.viewOptions.list_columnSizing) {
-			$tr.find('th').each(function (i, th) {
-				var $th = $(th);
-				var isLast = ($(this).next('th').length === 0);
-
+		var i, l, newWidth, taken;
+
+		if (this.viewOptions.list_columnSizing) {
+			i = 0;
+			taken = 0;
```
futuremint: I _think_ this means "widthTaken" or something like that?  But again... this needs a better name... its ok for the name to be longer, it will make the code clearer when the next person has to come along here and fix bugs/make changes.

js/repeater-list.js
```diff
@@ -823,57 +791,34 @@
...
-
-		}
-		return false;
-	};
-
-	var renderThead = function renderThead ($table, data) {
+	function renderThead ($table, data) {
 		var columns = data.columns || [];
 		var $thead = $table.find('thead');
+		var i, j, l, $tr;
```
futuremint: Pick better names for these index variables, or just use `.each`.  the letters `i` and `j` look too similar making this code hard to read and easy to make errors when changing it.

PR #1627, opened 2015-11-23
Scheduler: Weekdays frequency should be 'weekly' instead of 'daily'

nmerchant:
When selecting 'Weekdays' as the repeat option of the scheduler control, the recurrencePattern is as follows:

`FREQ=DAILY;BYDAY=MO,TU,WE,TH,FR;INTERVAL=1`

when it should be:

`FREQ=WEEKLY;BYDAY=MO,TU,WE,TH,FR;INTERVAL=1`

This can be verified against the recurrencePattern that gets set if you choose 'Weekly' as the repeat option, and manually select each week day, as these are functionally identical.

### DIFF
--- a/js/scheduler.js
+++ b/js/scheduler.js
@@ -554,10 +554,14 @@
 				}
 
 				if (recur.FREQ === 'DAILY') {
-					if (recur.INTERVAL === '1' && recur.COUNT === '1') {
-						item = 'none';
+					if (recur.BYDAY === 'MO,TU,WE,TH,FR') {
+							item = 'weekdays';
 					} else {
-						item = 'daily';
+						if (recur.INTERVAL === '1' && recur.COUNT === '1') {
+							item = 'none';
+						} else {
+							item = 'daily';
+						}
 					}
 				} else if (recur.FREQ === 'SECONDLY') {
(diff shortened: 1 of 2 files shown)


js/scheduler.js
```diff
@@ -554,14 +554,10 @@
 				}
 
 				if (recur.FREQ === 'DAILY') {
-					if (recur.BYDAY === 'MO,TU,WE,TH,FR') {
```
interactivellama: Please allow "old API" to work here. We can add a comment:

```
// Weekdays should be set farther along in this conditional when frequency is `WEEKLY`, but in order to maintain backwards compatibility with an older version of `setValue` the following few lines set the UI to weekdays when frequency is `DAILY` also.
```

interactivellama:
Please note this could be a breaking change if you are expecting the old `FREQ=DAILY;BYDAY=MO,TU,WE,TH,FR;INTERVAL=1` when the Repeat: Weekdays is selected.

interactivellama:
@dkilgore Does this change of output RRULE string cause any issue with your team?

kevinparkerson:
All checks out! Thanks @nmerchant

PR #1628, opened 2015-11-23
testing against bundle and min

swilliamset:


kevinparkerson:
ah, yes that makes more sense for dist tests doesn't it :D

PR #1635, opened 2015-11-24
Increase readability and maintainability: scheduler setValue

interactivellama:
So that it doesn't change from the values set via `setValue`.

This means that the time offset is ignored in the `startDateTime` option.

### DIFF
--- a/js/scheduler.js
+++ b/js/scheduler.js
@@ -199,33 +199,33 @@
 			this.toggleState('enable');
 		},
 
-		setUtcTime: function setUtcTime(d, t, offset) {
-			var date = d.split('-');
-			var time = t.split(':');
+		setUtcTime: function setUtcTime(day, time, offset) {
+			var dateSplit = day.split('-');
+			var timeSplit = time.split(':');
 			function z(n) {
 				return (n < 10 ? '0' : '') + n;
 			}
 
-			var utcDate = new Date(Date.UTC(date[0], (date[1] - 1), date[2], time[0], time[1], (time[2] ? time[2] : 0)));
+			var utcDate = new Date(Date.UTC(dateSplit[0], (dateSplit[1] - 1), dateSplit[2], timeSplit[0], timeSplit[1], (timeSplit[2] ? timeSplit[2] : 0)));
 
 			if (offset === 'Z') {
 				utcDate.setUTCHours(utcDate.getUTCHours() + 0);
 			} else {
-				var re1 = '(.)';// Any Single Character 1
-				var re2 = '.*?';// Non-greedy match on filler
-				var re3 = '\\d';// Uninteresting: d
-				var re4 = '.*?';// Non-greedy match on filler
-				var re5 = '(\\d)';// Any Single Digit 1
+				var expression = [];
+				expression[0] = '(.)'; // Any Single Character 1
+				expression[1] = '.*?'; // Non-greedy match on filler
+				expression[2] = '\\d'; // Uninteresting and ignored: d
+				expression[3] = '.*?'; // Non-greedy match on filler
+				expression[4] = '(\\d)'; // Any Single Digit 1
 
-				var p = new RegExp(re1 + re2 + re3 + re4 + re5, ["i"]);
-				var m = p.exec(offset);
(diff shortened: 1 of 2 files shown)


swilliamset:
@interactivellama please look into the CI fail

swilliamset:
image

interactivellama:
I couldn't allow the trainwreck that is `setValue` to remain. Not only didn't the function use a variable named `temp`, it also re-used the variable multiple times _and_ re-assigned values via array indexes when it was currently an array. I felt like I was back in "we-have-to-make-this-work-in-640k-memory-land."

js/scheduler.js
```diff
@@ -199,35 +199,17 @@
 			this.toggleState('enable');
 		},
 
-		setUtcTime: function setUtcTime(d, t, offset) {
-			var date = d.split('-');
-			var time = t.split(':');
+		setUtcTime: function setUtcTime(day, time, offset) {
+			var dateArray = day.split('-');
```
swilliamset: change variable name to `dateSplit` and `timeSplit` for consistency with other split results below. why not `*Array`? personal preference. `varnameArray[0]` feels redundant.

js/scheduler.js
```diff
@@ -199,35 +199,17 @@
...
 			function z(n) {
 				return (n < 10 ? '0' : '') + n;
 			}
 
-			var utcDate = new Date(Date.UTC(date[0], (date[1] - 1), date[2], time[0], time[1], (time[2] ? time[2] : 0)));
+			var utcDate = new Date(Date.UTC(dateArray[0], (dateArray[1] - 1), dateArray[2], timeArray[0], timeArray[1], (timeArray[2] ? timeArray[2] : 0)));
 
 			if (offset === 'Z') {
 				utcDate.setUTCHours(utcDate.getUTCHours() + 0);
-			} else {
```
swilliamset: How are we getting away with removing all of this?

interactivellama: I've only confirmed that it does work with it out and doesn't with it. I'm not sure why it is present. It's supposed to add an offset based on the timezone, but it's like it adds a second offset, or it could be this was designed to work without timezones specified in the input

js/scheduler.js
```diff
@@ -477,200 +459,239 @@
...
+			var $repeatMonthlyDate, $repeatYearlyDate, $repeatYearlyDay;
 
-					minutes = (minutes < 10) ? '0' + minutes : minutes;
-					startTime = hours + ':' + minutes;
-					temp = hours + ':' + minutes + ' ' + period;
-					this.$startTime.find('input').val(temp);
-					this.$startTime.combobox('selectByText', temp);
-				} else {
-					startTime = '00:00';
+			var semiColonSplitPattern = recurrencePattern.toUpperCase().split(';');
```
swilliamset: `semiColonPatternSplit` again for consistency with return values of `split()`

js/scheduler.js
```diff
@@ -477,200 +459,239 @@
...
+			for (i = 0; i < semiColonSplitPattern.length; i++) {
+				if (semiColonSplitPattern[i] !== '') {
+					item = semiColonSplitPattern[i].split('=');
```
swilliamset: and here

swilliamset:
Unable to determine if https://github.com/interactivellama/fuelux/commit/2979e6067733cb986059380e84b6df56b1ea86a2is the solution or just happy coincidence. Without a more complete range of unit tests the side affects are unknown. Please provide an assessment of the effort required to improve unit tests.

interactivellama:
Will remove actual "bugfix" and submit the refactor part and open a new PR with the "bugfix" change.

interactivellama:
`setUTCtime` was added at this point. I will follow up with @BenjaminNeilDavis for background when he becomes available. https://github.com/ExactTarget/fuelux/commit/008788b36df8d09eac61d08a7a42753caea81943

interactivellama:
Updated with refactor only. "Bugfix" removed.

swilliamset:
still need unit tests

swilliamset:
@dwaltz confirmed unit tests cover is good

PR #1645, opened 2015-12-03
separates out input-group-addon styles and updates index page html

### LINKED ISSUE #1644: Break input-group-addon styles out from checkbox-inline
We are re-using some of the checkbox-inline styles to make the input-group-addon look roughly how we want it in Fuel UX proper. But, in the mc-theme this causes it to break completely. Separating them allows us more fine-grained control over the style and addresses the issues it is causing in the mc-theme.

This is an addition of new style definitions and a change in the recommended markup. Those using the old markup will still be able to do so without any repercussions.

related to: https://github.exacttarget.com/uxarchitecture/fuelux-site/issues/4

cormacmccarthy:
Chrome:
image

IE10
image

IE11
image

Firefox
image

fixes #1644 

Addresses https://github.exacttarget.com/uxarchitecture/fuelux-site/issues/4

### DIFF
--- a/index.html
+++ b/index.html
@@ -144,6 +144,26 @@
 						</div>
 					</section>
 
+					<section id="checkbox-add-on" class="x-panel x-panel-default">
+						<header class="hidden x-panel-heading">
+							<h3 class="panel-title">Checkbox within input-group-addon</h3>
+						</header>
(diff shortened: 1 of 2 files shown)


kevinparkerson:
Looks good!

PR #1646, opened 2015-12-03
Placard contenteditable div support

### LINKED ISSUE #1612: Make placard support non-input/textarea this.$field elements
The placard should be able to support general elements (content-editable and otherwise) as the `.placard-field`. This would allow more extensive content to be edited via a placard

kevinparkerson:
Fixes #1612 by allowing contenteditable divs to be used as the `.placard-field` element in placard. I've made it default to look and behave like a regular input, though you can add `data-textarea="true"` to the `.placard-field` element to make it instead look and behave like a textarea.

Additionally, I fixed a bug I noticed with ellipsis for textareas.

Documentation will need to be updated separately.

### DIFF
--- a/index.html
+++ b/index.html
@@ -719,6 +719,34 @@
 							</div>
 						</div>
 					</div>
+
+					<br/>
+
+					<div class="placard" data-ellipsis="true" data-initialize="placard" id="myPlacard4">
+						<div class="placard-popup"></div>
+						<div class="placard-header">
+							<h5>Header</h5>
+						</div>
+						<div class="form-control placard-field" contenteditable="true"></div>
+						<div class="placard-footer">
+							<a class="placard-cancel" href="#">Cancel</a>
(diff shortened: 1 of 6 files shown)


interactivellama:
I think the logic is sound. Can you change `isDiv` to something more descriptive like `isContentEditableDiv`? And should we check that's it's content editable when create `isDiv`?
I'm coming from the greenfield of Facades, so any additional comments (especially in the constructor) would be nice. 

_(a keyCode object map would be nice for readability, too, but that's not what you are in there to add)._

kevinparkerson:
@interactivellama I can change the var name to `isContentEditableDiv`, though it shouldn't check if it is content editable, b/c if it is disabled on init `contenteditable="true"` wouldn't be present and the control would think it's a textarea

interactivellama:
Yeah, you're right. Let's just comment the constructor. 
`// Placard supports four types of input fields. A <div> that is content editable or an <input>. Either can function as a text field or a text area.`

PR #1672, opened 2016-01-12
Datepicker shows selected date when opened

### LINKED ISSUE #1671: Datepicker needs to revert to selected position on close if a selection has been made
self-explanatory :) makes it match fuelux v1 behavior

kevinparkerson:
fixes #1671 by having the calendar with the selected date visible when the datepicker is shown. If no selected date is present today's date is used.

Also adds show and hide methods, along with 'shown' and 'hidden' events. These will need to be documented.

Unit tests have been added to accompany these additions.

I've funneled all entry and exit points through the show and hide methods. I wanted to make use of `show.bs.dropdown` instead of `shown.bs.dropdown` (in order to switch the calendar screen before the dropdown was made visible) but could not due to a bug with Bootstrap's dropdown (https://github.com/twbs/bootstrap/blob/master/js/dropdown.js#L80 should be `return false;`). That bug is unlikely to be fixed, so `shown.bs.dropdown` will have to do. There is no visible impact anyway since the calendar screen renders very quickly (it's the same thing used when switching months)

### DIFF
--- a/js/datepicker.js
+++ b/js/datepicker.js
@@ -70,6 +70,7 @@
 		this.$header = this.$calendar.find('.datepicker-calendar-header');
 		this.$headerTitle = this.$header.find('.title');
 		this.$input = this.$element.find('input');
+		this.$inputGroupBtn = this.$element.find('.input-group-btn');
 		this.$wheels = this.$element.find('.datepicker-wheels');
 		this.$wheelsMonth = this.$element.find('.datepicker-wheels-month');
 		this.$wheelsYear = this.$element.find('.datepicker-wheels-year');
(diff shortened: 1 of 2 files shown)


js/datepicker.js
```diff
@@ -271,14 +274,26 @@
 			}
 		},
 
-		showDropdown: function (e) {
-			if (!this.$input.is(':focus')){
-				this.$element.find('.input-group-btn').addClass('open');
+		show: function () {
+			var date = (this.selectedDate) ? this.selectedDate : new Date();
```
swilliamset: `this.selectedDate || new Date()` Or Gate

swilliamset:
see https://github.com/ExactTarget/fuelux/issues/1671#issuecomment-175112271 for "enhancement" label removal

PR #1674, opened 2016-01-14
adds action option for list_selectable

jelzimme:
Introduces an additional option to the repeater list view that allows for an action column and header action cell without selectable rows or checkboxes. In this case, the header action cell affects every visible row.
This is triggered by using list_selectable = 'action'

### DIFF
--- a/js/repeater-list.js
+++ b/js/repeater-list.js
@@ -259,10 +259,13 @@
 				$actionsColumn.find('tr td:not(:last-child)').remove();
 
 				// Dont show actions dropdown in header if not multi select
-				if (this.viewOptions.list_selectable === 'multi') {
+				if (this.viewOptions.list_selectable === 'multi' || this.viewOptions.list_selectable === 'action') {
 					$actionsColumn.find('thead tr').html('<th><div class="repeater-list-heading">' + selectlist + '</div></th>');
-					//disable the header dropdown until an item is selected
-					$actionsColumn.find('thead .btn').attr('disabled', 'disabled');
+
+					if (this.viewOptions.list_selectable !== 'action') {
+						//disable the header dropdown until an item is selected
+						$actionsColumn.find('thead .btn').attr('disabled', 'disabled');
+					}
 				}
 				else {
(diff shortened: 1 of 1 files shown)


swilliamset:
@interactivellama, i've been involved in reviewing this pattern with @jamin-hall and have done a preliminary code review. could you do the merge after a secondary code review? good to have two sets of eyes when dealing with repeater.

interactivellama:
I've been unable to get to work as intended by modifying the "actions repeater" example on the index.html page. I easily may need to code it differently.

```
		var repeaterActions = $('#myRepeaterActions');
		repeaterActions.repeater({
			list_noItemsHTML: '<span>foo</span>',
			list_highlightSortedColumn: true,
			list_selectable: 'multi',
			list_actions: {
				width: 37,
				items: [
					{
						name: 'edit',
						html: '<span class="glyphicon glyphicon-pencil"></span> Edit'
					},
					{
						name: 'copy',
						html: '<span class="glyphicon glyphicon-copy"></span> Copy'
					},
					{
						name: 'delete',
						html: '<span class="glyphicon glyphicon-trash"></span> Delete',
						clickAction: function (helpers, callback, e) {
							console.log('hey it worked');
							console.log(helpers);
							e.preventDefault();
							callback();
						}
					}
				]
			},
			// setup your custom datasource to handle data retrieval;
			// responsible for any paging, sorting, filtering, searching logic
			dataSource: getSampleDataSet
		});
	}
	initRepeaterActions();
```

jelzimme:
Working to resolve the inconsistency between the standard example and my test case.

jelzimme:
I was able to make another change that should resolve the inconsistencies. I did not commit a change to index.js as I wasn't sure that this would be part of the index page moving forward.

interactivellama:
Merging at @swilliamset direction.

swilliamset:
needs to be in a minor tag because of enhancement. reverting.

swilliamset:
remerged https://github.com/ExactTarget/fuelux/pull/1691

PR #1685, opened 2016-01-22
Add data to thumbnail element, return data on thumbnail_getSelectedItems

### LINKED ISSUE #1686: Thumbnail repeater extension does not return selected items' data
The repeater list extension returns data as expected, as an object containing the selected item's element as well as the item's data object.

The repeater thumbnail extension, however, only returns the item's element.

nmerchant:
Fixes #1686

### DIFF
--- a/js/repeater-thumbnail.js
+++ b/js/repeater-thumbnail.js
@@ -36,10 +36,7 @@
 		$.fn.repeater.Constructor.prototype.thumbnail_getSelectedItems = function () {
 			var selected = [];
 			this.$canvas.find('.repeater-thumbnail-cont .selectable.selected').each(function () {
-				selected.push( {
-					element: $(this)
-					, data: $(this).data('item_data')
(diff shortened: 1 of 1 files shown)


js/repeater-thumbnail.js
```diff
@@ -36,7 +36,10 @@
 		$.fn.repeater.Constructor.prototype.thumbnail_getSelectedItems = function () {
 			var selected = [];
 			this.$canvas.find('.repeater-thumbnail-cont .selectable.selected').each(function () {
-				selected.push($(this));
```
kevinparkerson: leave this line as-is, since data can be appended to the thumbnail item and accessed in that way

kevinparkerson:
looks good!

swilliamset:
removing "enhancement". bug fix to achieve parity with other `_getSelectedItems` functions in repeater

PR #1714, opened 2016-02-13
Add setMaxLimit & setMinLimit Methods to spinbox()

dhonions:
Two new methods for .spinbox()

.spinbox('setMinLimit',10)
.spinbox('setMaxLimit',99)


swilliamset:
call `_applyLimits` after setting values so spinbox will not be in invalid state

https://github.com/dhonions/fuelux/blob/master/dist/js/fuelux.js#L2696

PR #1720, opened 2016-02-18
Tree lazy loading

### LINKED ISSUE #1708: Tree lazy loading
- [x] Link at bottom of folder to "load more" (next page of).
- [x] exposing folder level paging in the data source
- [x] Show loader while loading
- [x] update documentation

image

cormacmccarthy:
closes #1708

Adds third leaf type to tree, `overflow`, that displays as a link wherever it appears in the tree, with whatever name the user provides as the link text. When clicked, behaves similar to clicking on a folder, except that it also passes its data attributes back to the datasource to allow the datasource to respond appropriately. Whatever is returned by the datasource is appended to the parent branch of the clicked overflow. Overflow is removed when it is clicked.

Incorporated this into the tree's handlebars template.

When adding tests I tried to filter by which items no longer had `data-template` and discovered that this attribute wasn't actually being removed (it seemed from the intent of the code that it should have been). I modified the tree code to remove this (it now runs `removeAttr("data-template")` after `removeData("template")` since it could feasibly cause further problems for others if they happen to call $() on a tree leaf (which would re-add the data back into jQuery's internal data store). This allowed for more fine-grained testing as well as potentially avoiding future bugs.

tree-overflow

### DIFF
--- a/js/tree.js
+++ b/js/tree.js
@@ -126,7 +126,7 @@
 			var $loader = $parent.find('.tree-loader:last');
 
 			if (isBackgroundProcess === false) {
-				$loader.removeClass('hide hidden'); // hide is deprecated
+				$loader.removeClass('hide hidden'); // jQuery deprecated hide in 3.0. Use hidden instead. Leaving hide here to support previous markup
 			}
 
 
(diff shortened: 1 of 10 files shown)


kevinparkerson:
from what I can tell this looks good

js/tree.js
```diff
@@ -89,28 +93,59 @@
...
+				}else{
+					$el.remove();
+				}
+			}
+
+			var $loader = $parent.find('.tree-loader:last');
+
 			if (isBackgroundProcess === false) {
-				loader.removeClass('hide hidden'); // hide is deprecated
+				$loader.removeClass('hide hidden'); // hide is deprecated
```
kevinparkerson: Change this comment to:
`//in jQuery >= v3.0 hide and show are changing behaviors`

PR #1721, opened 2016-02-18
Adds proper form wrapping and many missing attributes from form controls

### LINKED ISSUE #1716: Handlebars form element templates should contain form-group wrappers
As the underscore templates do.

The wrappers are crucial to proper display of form elements. They are extremely poorly documented and easy to forget. The templates should contain and show them by default.

We can consider adding some sort of `unwrapped` flag that allows us to add the wrapper in an `{{#unless unwrapped}}` if for some reason developers want their templates without the wrappers...

cormacmccarthy:
Closes #1716 

Form controls were missing form-group wrappers, which is a crucial component of the markup when actually trying to use the control in the context of a form (which should be a majority of use-cases).

While I was adding this I went through and added many of the missing attributes and options that were previously left out due to time constraints (when the templates were created, the decision was made to only add the attributes and options absolutely necessary for displaying the examples on the fuelux site).


swilliamset:
did we decide against the `{{#unless inline}} ` option

swilliamset:
never mind. the conversation before was about an `unwrapped` option which we shouldn't be encouraging. moving on.

PR #1723, opened 2016-02-19
Input elements should support validation and feedback

cormacmccarthy:
related to #1716 

image

Adds support for validation to all form templates which rely on the input element.

Adds support for feedback to input template.

Adds preliminary, but untested, feedback support to other templates which rely on input (search, combobox, etc. I added it while I was in there, but, it is undocumented and untested since it was not prioritized or necessary for the task at hand)


kevinparkerson:
Looked at this, checked out! :+1:

PR #1725, opened 2016-02-19
adds filter option to combobox and keyboard nav for dropdown

### LINKED ISSUE #1710: expose onkeyup in combobox for typeahead support
maybe use search input which has [onkeyup support](https://github.com/ExactTarget/fuelux/blob/master/js/search.js#L38-L40)

add documentation and example

cormacmccarthy:
closes #1710 

Allows the user to provide their own filter as an option if desired (or just use the basic built-in one).

By default, could be styled to support different displays for partial or full matches if desired.

The keyboard navigation also increases accessibility for the control.

Also fixes until-know-uncaught bug where `$(a).text()` contents may contain whitespace due to html formatting, and therefore `"One        "` and `"One"` would not match (now calling `.trim()` on `$(a).text()`)

Default filter:

- Styles exact matches as `.text-success`
- Styles matches that begin with input text as `.text-info`
- Hides non-matches
- Displays (non-selectable) "_No Matches_" when no matches are found

Combobox now:
- Allows navigation of options with arrow keys
- Changes selection to highlighted option on enter
- Clears combobox on `esc`
- Optionally fires filter on non-navigational keystrokes
- Fires change events on selection change through navigationally selected selections

combobox


js/combobox.js
```diff
@@ -232,7 +315,34 @@
 	};
 
 	$.fn.combobox.defaults = {
-		autoResizeMenu: true
+		autoResizeMenu: true,
+		filterOnKeypress: false,
+		showOptionsOnKeypress: false,
+		filter: function filter(list, predicate, self) {
```
kevinparkerson: please add a space between filter and (

js/combobox.js
```diff
@@ -232,7 +315,34 @@
...
+			var visible = 0;
+			self.$dropMenu.find('.empty-indicator').remove();
+
+			list.each(function(i){
```
kevinparkerson: same here, add a space between function and (, and ) and {

kevinparkerson:
only some minor formatting changes needed, once those are done this is good to go! :+1:

kevinparkerson:
Ah yes, those unit tests are nice! Good work! :+1:

PR #1764, opened 2016-03-16
Adds ability to cancel repeater search

### LINKED ISSUE #1733: Add ability to cancel repeater search
Add ability for user to cancel an in-progress search on the repeater.

Possibly fire a "searchCancelled" event. Make sure that if the long-running search does eventually execute successfully there is a way to keep it from clobbering the repeater values/display.

Requested by Brian Tuttle for email team.

scott-joe:
Fixes #1733 

- Added `pending` state to Search
- Added `allowsCancel` option to Search. Defaults to `FALSE`
- Repeater initializes Search with `allowsCancel: TRUE`
- On search, Search issues `searchChange` event to Repeater. Repeater disables Search, which with the `allowsCancel` option, does not disable the button. When Repeater is finished rendering, the `rendered` event is pushed to Search, removing the `pending` status and class. This returns the button to Clear functionality. If the button is pushed during `pending` status, the `canceled` event is pushed to Repeater. Repeater receives this event, and pushes it's own event out for the datasource to handle the cancelation of the XHR request or simply ignoring the results. Meanwhile, a render is performed and Search is cleared as normal.
- Attempted to write a unit test for this, but the existing unit tests don't seem to perform all 3 cases, so I'll need advising on how this was intended to work, and how I can fix both perhaps.
- Docs PR to follow after solution is merged

### DIFF
--- a/js/repeater.js
+++ b/js/repeater.js
@@ -76,7 +76,8 @@
 		this.$pageSize.selectlist();
 		this.$primaryPaging.find('.combobox').combobox();
 		this.$search.search({
-			searchOnKeyPress: this.options.searchOnKeyPress
+			searchOnKeyPress: this.options.searchOnKeyPress,
+			allowCancel: true
 		});
 
 		this.$filters.on('changed.fu.selectlist', function (e, value) {
@@ -105,6 +106,10 @@
 				pageIncrement: null
 			});
 		});
+		this.$search.on('canceled.fu.search', function (e, value) {
+			self.$element.trigger('canceled.fu.repeater', value);
+		});
+
 		this.$secondaryPaging.on('blur.fu.repeater', function (e) {
 			self.pageInputChange(self.$secondaryPaging.val());
 		});
@@ -633,11 +638,13 @@
 					self.$loader.hide().loader('pause');
 					self.enable();
 
+					self.$search.trigger('rendered.fu.repeater');
 					self.$element.trigger('rendered.fu.repeater', {
 						data: data,
 						options: dataOptions,
 						renderOptions: options
 					});
+
 					//for maintaining support of 'loaded' event
 					self.$element.trigger('loaded.fu.repeater', dataOptions);
 				});
--- a/js/repeater.js
+++ b/js/repeater.js
@@ -77,7 +77,7 @@
 		this.$primaryPaging.find('.combobox').combobox();
 		this.$search.search({
 			searchOnKeyPress: this.options.searchOnKeyPress,
-			allowCancel: true
+			allowCancel: this.options.allowCancel
 		});
 
 		this.$filters.on('changed.fu.selectlist', function (e, value) {
@@ -108,6 +108,10 @@
 		});
 		this.$search.on('canceled.fu.search', function (e, value) {
 			self.$element.trigger('canceled.fu.repeater', value);
+			self.render({
+				clearInfinite: true,
+				pageIncrement: null
+			});
 		});
 
 		this.$secondaryPaging.on('blur.fu.repeater', function (e) {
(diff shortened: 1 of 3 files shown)


js/repeater.js
```diff
@@ -76,7 +76,8 @@
 		this.$pageSize.selectlist();
 		this.$primaryPaging.find('.combobox').combobox();
 		this.$search.search({
-			searchOnKeyPress: this.options.searchOnKeyPress
+			searchOnKeyPress: this.options.searchOnKeyPress,
+			allowCancel: true
```
futuremint: This means that the search in the repeater will _always_ be cancelable.  Should we also support overriding this option on repeater for someone who does _not_ want repeater searches to be cancelable?

js/search.js
```diff
@@ -73,7 +77,7 @@
 			}
 
 			this.activeSearch = searchText;
-			this.$element.addClass('searched');
+			this.$element.addClass('searched pending');
```
futuremint: Would it be better to use a variable on `this` instead of storing this state in a class name?

js/search.js
```diff
@@ -102,7 +115,10 @@
 			e.preventDefault();
 			if ($(e.currentTarget).is('.disabled, :disabled')) return;
 
-			if(this.$element.hasClass('searched')) {
+			if(this.$element.hasClass('pending')) {
```
futuremint: Instead of an if/else here, you can just combine them into a single `||` statement in the if expression.

test/repeater-test.js
```diff
@@ -166,6 +166,36 @@ define(function(require){
...
+		$repeater.repeater({
+			dataSource: function(options, callback){
+				count++;
+				switch (count){
+					case 0:
+						equal(options.search, undefined, 'search value not passed to dataSource initially as expected');
+						callback({});
+						$search.find('input').val('something');
+						$search.trigger('searched.fu.repeater');
+						break;
```
futuremint: Yep, this definitely doesn't work. I think I remembered figuring out why at some point in the past and it affected all these tests.  Partially its that the tests needs to run async using https://api.qunitjs.com/async/ and not to signal the test is over until all assertions have completed. Because triggering these DOM events is not guaranteed to be synchronous operation in every browser. I tried adding an async in here and the 2nd & 3rd branches still don't work, I'll mess with it more tomorrow morning.

js/repeater.js
scott-joe: The tech specs didn't cover whether it should always be cancelable, so allowing cancels seemed the less dangerous assumption. I can definitely allow passthrough for Repeater if you'd like it to.

js/search.js
scott-joe: Thats actually how I wrote it when we had our 1:1. If you'll remember, I had a mini-rant about storing state in classnames, but decided to follow convention. If I can't re-write it all, I honestly don't care either way, so if anyone has a preference, I can follow popular opinion.

js/search.js
scott-joe: Good call

test/repeater-test.js
scott-joe: Cool. I started to write a while around it, but I wasn't sure if the unit tests had something like that built in. I'll check out that documentation tonight.

js/repeater.js
```diff
@@ -105,6 +106,10 @@
 				pageIncrement: null
 			});
 		});
+		this.$search.on('canceled.fu.search', function (e, value) {
+			self.$element.trigger('canceled.fu.repeater', value);
```
futuremint: As discussed in the sprint retrospective, this also needs to re-render the repeater (see the handler above for how to call render). Keep in mind that render will ignore "re-renders" if its in the middle of a render, so the render method itself will need to be updated to allow stopping and re-rendering a render that it mid-render (which is the case here during a cancelled search).

js/repeater.js
futuremint: Its fine for now, we'll add it later if someone asks for it to be disabled.

scott-joe:
Made the requested changes.

PR #1773, opened 2016-03-22
Update package.json & travis to use node 4.4.x

futuremint:


futuremint:
@chriscorwin I pushed the changes from here to fuelux-edge so that Travis pulls the correct node_modules zip file.  Looks Travis passed...

PR #1776, opened 2016-03-23
Gh1768   add ability to text to automatically truncate when space requires

### LINKED ISSUE #1768: Add ability to text to automatically truncate when space requires
Add ability to truncate text.

Should follow the example of the [SLDS Truncation Utility](http://www.lightningdesignsystem.com/components/utilities/truncation/).

slds-truncate-example

Now: do we want to make this an actual _control_ (or utility) that a developer instantiates and automatically puts the title attribute into place from the content? 

(I feel like that could get tricky if the content is anything except for plain text.)

Or do we just want to provide the CSS and require the developer put the title on manually?

chriscorwin:
Adds truncate mixin and helper class.

Also adds some help to individual controls styles where truncate might be applied to wrapper (like tree) or with form elements to compensate for the inline-block that truncate requires (like checkbox, radio).

Fixes #1768


grunt/config/connect.js
```diff
@@ -7,7 +7,7 @@ module.exports = function (grunt) {
 				base: {
 					path: '.',
 					options: {
-						index: ['index.html', 'tests.html']
+						index: ['index.html', 'index-dev.html', 'tests.html']
```
futuremint: Please remote index-dev.html from this file (or just revert the file).

PR #1812, opened 2016-05-17
Update tree markup to have alert role on span

interactivellama:
grunt serve task was failing do to html lint errors. Apparently, you can't have alert ARIA roles on li tags. They need to be on the ul or something inside the li tag.

### DIFF
--- a/markup/tree-folders-selectable.html
+++ b/markup/tree-folders-selectable.html
@@ -12,7 +12,7 @@
 			</button>
 		</div>
 		<ul class="tree-branch-children" role="group"></ul>
-		<div class="tree-loader" role="alert">Loading...</div>
+		<div class="tree-loader"><span role="alert">Loading...</span></div>
 	</li>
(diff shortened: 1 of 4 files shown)


interactivellama:
This was erroring out for me when doing repeater/combobox fixes.

markup/tree-folders-selectable.html
```diff
@@ -12,7 +12,7 @@
 			</button>
 		</div>
 		<ul class="tree-branch-children" role="group"></ul>
-		<div class="tree-loader" role="alert">Loading...</div>
+		<div class="tree-loader"><span role="alert">Loading...</span></div>
```
futuremint: You said "you can't have `role="alert"` on li tags" but there are also a lot of changes here on div tags too.  Are you fixing more than necessary here?

interactivellama: I was trying to make it consistent and always you use a `span`. You are right the `div`'s don't need to change to pass.

PR #1816, opened 2016-05-23
Repeater combobox events

### LINKED ISSUE #1790: Repeater: pageChanged.fu.repeater fires itself twice on every pagechange
Bind the pageChanged.fu.repeater to your repeater, in my example it is spitting out a console.log. See that the event is fired twice when the user changes the page number in the input box.

Expected Result: this event is fired once.

If this worked as expected it would facilitate a way to limit the user from entering page numbers out of the range provided by the datasource.

CodePen example: http://codepen.io/dkilgore/pen/oxpNzj

interactivellama:
Fixes #1790 for @dkilgore 

`pageChanged.fu.repeater` should now only fire once. Repeater's `pageChanged` event will now return an array with a `[ [pageNumber], [dataObject] ]` shape.

**Additional help with paging and multiple `combobox` events**
If repeater only has two pages of data, the combobox will be present. If a `3` is entered into the combobox the data key will return the data from the combobox's `changed` event 

`{ notFound: true, text: '3'}`

The `3` is not present (since there are only two pages), therefore a `notFound` will be present in the data returned from the `combobox`. A 3 (or really a zero index 2, will still be passed into the repeater, but you can now check for the `notFound` key in order to revert to a valid page if you so choose.

Combobox was refactored, so that that `changed.fu.combobox` event did not fire twice when changed after typing and then using the enter key. Issues were also present that created multiple selected menu items due to previous item with `.selected` class were not being removed, but still determining what item selection by a query selector for `.selected:first`.

**Individual tests**
Individual tests for combobox were not being "taken down" and therefore tests were affecting each other and not isolated. It's a `noop/return` if you run `.combobox()` on a DOM element that has already been initialized.

_Please note: There are combobox tests that test events fire once, but there are no tests to check that `pagechanged.fu.repeater` runs once or any tests related to that event at all._

### DIFF
--- a/js/combobox.js
+++ b/js/combobox.js
@@ -83,12 +83,22 @@
 		},
 
 		doSelect: function ($item) {
+			
 			if (typeof $item[0] !== 'undefined') {
-				$item.addClass('selected');
+				// remove selection from old item, may result in remove and 
+				// re-addition of class if item is the same
+				this.$element.find('li.selected:first').removeClass('selected');
+
+				// add selection to new item
 				this.$selectedItem = $item;
+				this.$selectedItem.addClass('selected');
+
+				// update input
 				this.$input.val(this.$selectedItem.text().trim());
 			} else {
+				// this is a custom input, not in the menu
 				this.$selectedItem = null;
+				this.$element.find('li.selected:first').removeClass('selected');
 			}
 		},
 
@@ -120,7 +130,8 @@
 				}, this.$selectedItem.data());
 			} else {
 				data = {
-					text: this.$input.val().trim()
+					text: this.$input.val().trim(),
+					notFound: true
 				};
 			}
 
@@ -232,7 +243,6 @@
 				}
 
 				this.$inputGroupBtn.removeClass('open');
-				this.inputchanged(e);
 			} else if (e.which === ESC) {
 				e.preventDefault();
 				this.clearSelection();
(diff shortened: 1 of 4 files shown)


js/combobox.js
```diff
@@ -270,10 +279,13 @@
 		},
 
 		inputchanged: function (e, extra) {
+			var val = $(e.target).val();
```
futuremint: is there a reason you moved this above the early return?

js/combobox.js
```diff
@@ -270,10 +279,13 @@
...
 			// skip processing for internally-generated synthetic event
 			// to avoid double processing
-			if (extra && extra.synthetic) return;
-			var val = $(e.target).val();
+			if (extra && extra.synthetic) {
+				this.selectByText(val);
```
futuremint: Does the addition of calling selectByText here help fix this bug?

interactivellama: If the synthetic event of the menu click happens. The menu item never gets the `.selected` class.

js/combobox.js
interactivellama: It's so that `val` is defined.

js/repeater.js
```diff
@@ -96,8 +96,7 @@
 		});
 		this.$prevBtn.on('click.fu.repeater', $.proxy(this.previous, this));
 		this.$primaryPaging.find('.combobox').on('changed.fu.combobox', function (evt, data) {
```
interactivellama: @futuremint This event gets fired twice in certain cases with combobox which then calls `pageInputChange` twice.

PR #1847, opened 2016-07-21
Update repeater header checkbox to use correct markup

interactivellama:
Hello, Fuel UX, old friend. It's been a while.

Partial fix to #251 in GHE's Fuel UX site. This corrects the markup generated by the repeater.

There shouldn't be any visual changes here. The main point was to wrap the `input` in the `label` to align with markup found in the [Fuel UX Checkbox](http://getfuelux.com/javascript.html#checkbox-examples-inline), so that we have the same starting point for checkboxes for the Lightning Theme.

screen shot 2016-07-21 at 12 19 37 am

### DIFF
--- a/js/repeater-list.js
+++ b/js/repeater-list.js
@@ -404,6 +404,7 @@
 
 		$.fn.repeater.Constructor.prototype.list_frozenOptionsInitialize = function () {
 			var $checkboxes = this.$element.find('.frozen-column-wrapper .checkbox-inline');
+			var $headerCheckbox = this.$element.find('.header-checkbox .checkbox-custom');
 			var $everyTable = this.$element.find('.repeater-list table');
 			var self = this;
 
@@ -419,6 +420,7 @@
 				}
 			});
 
+			$headerCheckbox.checkbox();
 			$checkboxes.checkbox();
 
 			// Row checkboxes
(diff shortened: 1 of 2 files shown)


js/repeater-list.js
```diff
@@ -687,8 +689,14 @@
 		var chevUp = 'glyphicon-chevron-up';
 		var $div = $('<div class="repeater-list-heading"><span class="glyphicon rlc"></span></div>');
 		var checkAllID = (this.$element.attr('id')+'_' || '') + 'checkall';
-		var checkBoxMarkup = '<div class="repeater-list-heading header-checkbox"><div class="checkbox checkbox-inline"><input type="checkbox" id="' + checkAllID + '">'+
-			'<label for="' + checkAllID + '"></label></div></div>';
+
+		var checkBoxMarkup = '<div class="repeater-list-heading header-checkbox">' +
+				'<label for="' + checkAllID + '" class="checkbox-custom checkbox-inline">' + 
+					'<input id="' + checkAllID + '" class="sr-only" type="checkbox" value="">' +
```
cormacmccarthy: According to the markup folder, the `id` goes on the label, not the input.

interactivellama: Sounds good. I knew that but was trying to make I didn't break things. Will update and test.

interactivellama: Yep, it doesn't work when the `id` is on the label, I'm guessing that the repeater is interacting with the checkbox directly.

js/repeater-list.js
```diff
@@ -687,8 +689,14 @@
 		var chevUp = 'glyphicon-chevron-up';
 		var $div = $('<div class="repeater-list-heading"><span class="glyphicon rlc"></span></div>');
 		var checkAllID = (this.$element.attr('id')+'_' || '') + 'checkall';
-		var checkBoxMarkup = '<div class="repeater-list-heading header-checkbox"><div class="checkbox checkbox-inline"><input type="checkbox" id="' + checkAllID + '">'+
-			'<label for="' + checkAllID + '"></label></div></div>';
+
+		var checkBoxMarkup = '<div class="repeater-list-heading header-checkbox">' +
+				'<label id="' + checkAllID + '" class="checkbox-custom checkbox-inline">' +
```
interactivellama: @cormacmccarthy Updated.

PR #1850, opened 2016-08-04
Adds list_renderRowAfterColumns option

aheile:
Adds repeater-list option: list_renderRowAfterColumns to wait to append the
row until its columns have already been appended to it.
This helps with some Firefox unresponsive script warnings in trying to render our repeater.

### DIFF
--- a/js/repeater-list.js
+++ b/js/repeater-list.js
@@ -489,7 +489,8 @@
 			list_sortClearing: false,
 			list_rowRendered: null,
 			list_frozenColumns: 0,
-			list_actions: false
+			list_actions: false,
+			list_renderRowAfterColumns: false
 		});
 
 		//EXTENSION DEFINITION
@@ -856,10 +857,16 @@
 			$row.data('item_data', rows[index]);
 		}
 
-		$tbody.append($row);
+		if (!this.viewOptions.list_renderRowAfterColumns) {
+			$tbody.append($row);
+		}
 
 		for (i = 0, length = this.list_columns.length; i < length; i++) {
 			renderColumn.call(this, $row, rows, index, this.list_columns, i);
+		}
+
+		if (this.viewOptions.list_renderRowAfterColumns) {
+			$tbody.append($row);
 		}
 
 		if (this.viewOptions.list_rowRendered) {


mtrepina:
@interactivellama Any possibility of prioritizing this PR.  We are seeing some bad bad things with the repeater in firefox when 500+ cells are rendered.

futuremint:
@mtrepina @swilliamset is handling Fuel UX requests now, please ask him to prioritize reviewing this in their next sprint.  Thanks.

futuremint:
Also, @aheile can you please add a unit test for this that validates this solves the slow script problem?

aheile:
Here is a sample with lots of cells which averages 13-15 seconds for me in Firefox:
http://jsfiddle.net/qpyygcqw/

Verses 7-8 seconds:
http://jsfiddle.net/mfy29f41/

cormacmccarthy:
@aheile what is the purpose of adding this as an option instead of just moving where the row gets appended to the body? Basically: what are the potential adverse effects of moving `$tbody.append($row);` below the `for` loop and calling it a day?

aheile:
@cormacmccarthy I would be fine with that. I was just trying to minimize any potential negative impact that I might not be aware of.

cormacmccarthy:
Reading through the code a bit further it does look like there may be some potential side effects of doing it this way. Specifically, right now `viewOptions.list_columnRendered` gets called each time a new cell is attached to a row which is currently already attached to the DOM. The new way would call `list_columnRendered` _before_ the row (and therefore the cell) is attached to the DOM. I believe that Automation Studio either was or is using this function to adjust the size of their cells (or something like that). So, just moving the append would break that (because you couldn't properly measure the cell yet if it isn't attached to the DOM).

However, instead of passing in an option that has to be documented, etc, we should perhaps rip this portion out of the `renderColumn` method :

		if (!(columns[columnIndex].property === '@_CHECKBOX_@' || columns[columnIndex].property === '@_ACTIONS_@') && this.viewOptions.list_columnRendered) {
			this.viewOptions.list_columnRendered({
				container: $row,
				columnAttr: columns[columnIndex].property,
				item: $col,
				rowData: rows[rowIndex]
			}, function () {});
		}

and, if it is going to be necessary to call it, call it column by column after the row is appended to the DOM, roughly like this (this probably won't work, will need some adjustments):

		for (i = 0, length = this.list_columns.length; i < length; i++) {
			renderColumn.call(this, $row, rows, index, this.list_columns, i);
		}

		$tbody.append($row);

		if (!(columns[columnIndex].property === '@_CHECKBOX_@' || columns[columnIndex].property === '@_ACTIONS_@') && this.viewOptions.list_columnRendered) {
			for (i = 0, length = this.list_columns.length; i < length; i++) {
				this.viewOptions.list_columnRendered({
					container: $row,
					columnAttr: columns[columnIndex].property,
					item: $col,
					rowData: rows[rowIndex]
				}, function () {});
			}
		}

cormacmccarthy:
Closing in favor of #1863

PR #1857, opened 2016-08-22
Changed the path to the latest version to the correct one.

Jandev:
Pointing to the correct zip-file. 
The current zip file points to an old version (3.4), which is a bit outdated.


cormacmccarthy:
Thanks!

PR #1863, opened 2016-09-08
Attaches row to DOM after all columns have been added

cormacmccarthy:
fixes #1850 

JIRA ticket: https://jira.exacttarget.com:8443/browse/FUELUX-252

Because the columns need to actually be attached to the DOM at the time `list_columnRendered` is called, I broke that portion out of the `renderColumn` function and placed it in its own separate loop (which is only run if there is actually a `list_columnRendered` function to be called). I believe Automation Studio in particular uses (or used) this function to measure the size of a cell and then do something with that information (can't remember what).

Makes the `repeater.js` and `repeater-list.js` code (and related test files) conform to our current eslint standards across Fuel UX properties to try and increase maintainability (since these two areas in the code in particular have been notoriously hard to comprehend and work with).

There's not really any good way to unit test this changes since it is merely a speed improvement. I tested how long it took the repeater to render in firefox and chrome before and after the change. You can see the results of running the test 10x in each below:

All results are reported in units of seconds.

In Chrome:
Without changes:
(0.827 + 0.729 + 0.779 + 0.668 + 0.686 + 0.689 + 0.683 + 0.744 + 0.677 + 0.698) / 10
0.718 seconds on average

With changes (did it twice because I was surprised it didn't really make a difference, second set was even worse somehow):
(0.688 + 0.759 + 0.694 + 0.709 + 0.691 + 0.738 + 0.671 + 0.812 + 0.724 + 0.676) / 10
0.716 average
(0.721 + 0.744 + 0.703 + 0.697 + 0.71 + 0.805 + 0.684 + 0.722 + 0.751 + 0.764) / 10
0.7301 seconds on average

In Firefox:
Without changes:
(0.979 + 0.925 + 0.911 + 0.922 + 0.901 + 0.915 + 0.889 + 0.938 + 0.891 + 0.964) / 10
0.9235 seconds on average

With changes:
(0.905 + 0.912 + 0.869 + 0.927 + 0.946 + 0.973 + 0.905 + 0.846 + 0.913 + 0.964) / 10
0.916 seconds on average

What I believe is happening here is that the Unit Tests run against a shadow DOM instead of an actual rendered DOM. Since this change only speeds up things if it is running against an actual rendered DOM the unit tests will not help us here unless they are changed to render the repeater out to the actual page. 

I created a test/repeater.html that would, using the data the costumer provided in #1850, render the repeater to the page and measure how long the render took. From the results you can see that there is a modest performance gain to be had with these changes (more pronounced in Firefox):

Rendered DOM
Chrome:
Without Changes:
(1.624 + 1.651 + 1.858 + 1.724 + 1.834 + 1.722 + 1.675 + 1.665 + 1.717 + 1.734) / 10
1.7204 seconds on average

With Changes:
(1.662 + 1.691 + 1.621 + 1.611 + 1.711 + 1.649 + 1.66 + 1.695 + 1.65 + 1.655) / 10
1.6605 seconds on average

Firefox:
Without Changes:
(1.887 + 1.686 + 1.796 + 1.703 + 1.796 + 1.77 + 1.817 + 1.797 + 1.818 + 1.784) / 10
1.7854 seconds on average

With Changes:
(1.535 + 1.442 + 1.43 + 1.436 + 1.496 + 1.505 + 1.437 + 1.48 + 1.426 + 1.45) / 10
1.4637 seconds on average

To test:
- [x] Check this branch out
- [x] Reset to commit `acbc3eb`
- [x] `grunt servefast`
- [x] Open [http://localhost:8000/test/repeater.html](http://localhost:8000/test/repeater.html)
- [x] Open the console and observe the time it took to render (displayed something like "1.785 + ")
- [x] Reset to commit `1c4bb0e`
- [x] `grunt servefast`
- [x] Open [http://localhost:8000/test/repeater.html](http://localhost:8000/test/repeater.html)
- [x] Open the console and observe the time it took to render (displayed something like "1.785 + ")
- [x] Confirm that the render is, on average, faster with the changes applied
- [x] Run unit tests
- [x] Review code changes


js/repeater.js
```diff
@@ -227,19 +312,17 @@
...
-			this.$prevBtn.attr(disabled, disabled);
-			this.$nextBtn.attr(disabled, disabled);
+			this.$search.search('disable');
+			this.$filters.selectlist('disable');
+			this.$views.find('label, input').addClass('disabled').attr('disabled', 'disabled');
+			this.$pageSize.selectlist('disable');
+			this.$primaryPaging.find('.combobox').combobox('disable');
+			this.$secondaryPaging.attr('disabled', 'disabled');
+			this.$prevBtn.attr('disabled', 'disabled');
+			this.$nextBtn.attr('disabled', 'disabled');
```
interactivellama: I think someone must have been worried about the extra keystrokes here. :-) Good call!

js/repeater.js
```diff
@@ -370,8 +455,8 @@
 				});
 				footer.hide();
 			} else {
-				cont = this.infiniteScrollingCont;
-				data = cont.data();
+				var cont = this.infiniteScrollingCont;
+				var data = cont.data();
```
interactivellama: These would be hoisted anyway.

js/repeater.js
```diff
@@ -550,42 +606,35 @@
...
 			this.render({
 				pageIncrement: this.pageIncrement
 			});
 		},
 
-		render: function (options) {
-			var self = this;
+		// This functions more as a "pre-render" than a true "render"
+		render: function render (opts) {
+			this.disable();
```
interactivellama: I usually think of statements after declarations, but I guess there could be enabled UI elements for that 10ms or so.

js/repeater.js
```diff
@@ -617,48 +666,34 @@
...
-					self.itemization(data);
-					self.pagination(data);
-				}
-
-				self.runRenderer(viewTypeObj, data, function () {
-					if (self.infiniteScrollingEnabled) {
-						if (viewChanged || options.clearInfinite) {
-							self.initInfiniteScrolling();
+			var beforeRender = this.viewOptions.dataSource;
+			var self = this;
```
interactivellama: I know this is standard, but something like `repeaterPrototypeContext` would be most clear.

js/repeater.js
```diff
@@ -691,90 +726,71 @@
...
 						container: this.$canvas,
 						data: data
 					});
-					addItem(this.$canvas, response);
+					addItem(this.$canvas, addBefore);
 				}
 
-				$container = this.$canvas.find('[data-container="true"]:last');
+				var $container = this.$canvas.find('[data-container="true"]:last');
 				$container = ($container.length > 0) ? $container : this.$canvas;
```
interactivellama: Should we do an if statement here, so we are self-assigning the same variable?

js/repeater.js
```diff
@@ -691,90 +726,71 @@
...
-						if (window.console && window.console.warn) {
-							window.console.warn('WARNING: Repeater plugin "repeat" value must start with either "data" or "this"');
-						}
+						logWarn('WARNING: Repeater plugin "repeat" value must start with either "data" or "this"');
 					}
 
-					for (i = 0, l = repeat.length; i < l; i++) {
-						if (subset[repeat[i]] !== undefined){
-							subset = subset[repeat[i]];
+					for (var repeatI = 0, repeatL = repeat.length; repeatI < repeatL; repeatI++) {
```
interactivellama: Can we just go ahead and spell these out?

js/repeater.js
```diff
@@ -691,90 +726,71 @@
...
 				}
 
-				$container = this.$canvas.find('[data-container="true"]:last');
+				var $container = this.$canvas.find('[data-container="true"]:last');
 				$container = ($container.length > 0) ? $container : this.$canvas;
 
 				if (viewTypeObj.renderItem) {
-					repeat = viewTypeObj.repeat || 'data.items';
+					var subset;
+					var repeat = viewTypeObj.repeat || 'data.items';
```
interactivellama: Could we do something like `viewTypeObjItems`. Not a big deal, but there's a lot going on and was struggling with the meaning of `repeat`

js/repeater.js
```diff
@@ -691,90 +726,71 @@
...
-							}
-							break;
+							logWarn('WARNING: Repeater unable to find property to iterate renderItem on.');
+							repeatI = repeat.length;
 						}
 					}
 
-					for (i = 0, l = subset.length; i < l; i++) {
-						response = viewTypeObj.renderItem.call(this, {
+					for (var subItemI = 0, subsetL = subset.length; subItemI < subsetL; subItemI++) {
```
interactivellama: Not completely sure but length may already by a cached value

http://stackoverflow.com/questions/5752906/is-reading-the-length-property-of-an-array-really-that-expensive-an-operation

interactivellama:
I can confirm about a half second improvement in FireFox.

interactivellama:
Looks good to me. There are so many variations for Repeater, it's difficult to review all of them, but I've read every line, confirmed that this is faster and was unable to find any bugs in the repeater examples I have.

js/repeater.js
cormacmccarthy: Yeah, I totally understand hoisting and understand the mindset behind declaring all variables at the point where they would be hoisted to anyways. The purpose for this change (and many others like it throughout this refactor) was keeping in line with "10.4 Scope" of Steve McConnell's **Code Complete**, I'm keeping the "live time" as short as possible (Live Time is the number of lines of code between which a variable is first and last referenced in the code). To summarize his arguments on pages 246/247, "A short live time reduces the chance of incorrectly or inadvertently altering a variable between the places in which you intend to alter it, gives you an accurate picture of your code's intention, reduces the chance of initialization errors, and makes your code more readable"

js/repeater.js
cormacmccarthy: `this.disable()` could theoretically get pushed further and further down as people add things to the code if they aren't paying attention. I put it at the very top, both as a signal to future developers that "the control is disabled while this function runs" and to prevent it getting accidentally pushed down too far (somehow) or lost entirely (very unlikely). I'm kind of treating it like an `e.preventDefault()` (which I would also put at the very top).

js/repeater.js
interactivellama: What do you think about declaring it at the, but assigning within the if?

js/repeater.js
cormacmccarthy: Man oh man, I went back and forth on restructuring this for clarity/readability as you suggest. Since jsperf was down and because of all the peculiarities of different browsers, and because of how freaking slow the repeater already is, I went with the "safe" (circa 2011) practice of manually caching.

However, re-examining this code this morning, a few things stand out to me:
1) This entire method is only called a _single time_
2) These `for` loops each only iterate a few times (the first one usually just _one time_ and the second only as many rows or thumbnails as are loaded on page, typically less than 100)

Not to mention the link you pointed to calls the entire practice into question (at least for Chrome).

Changing.

js/repeater.js
cormacmccarthy: Did you leave out the word "beginning"?

Live Time refers not to the variables life in the guts of the computer executing the code, but, rather to the variable's life in the brain of the programmer. Consider the following scenarios:

A) Programmer declares `var foo;` at the top of a long function. Sets `foo` to something near the middle of the function, uses it. Later comes back and sets `foo` to something else near the end of the function, forgetting that it was also used in the middle of the function. Bug.

B) Programmer declares and sets `foo` to something in the middle of the function. Later comes back and tries to declare and set `foo` near the end of the function, get's an eslint warning about `foo` already being declared. Names second `foo` something else. No bug.

Because of this, I believe variables should be declared as close to their first use as is practical/possible. Typically they should be declared at the moment they are set. 

And yes I understand hoisting blah blah blah. I have found and fixed several bugs as a result of people being terrified of hoisting or trying to be clever and showing they know about hoisting and declaring all their variables at the top of a function. It's dangerous and unnecessary. If the compiler does it anyways, we certainly don't need to. Especially when ESLint is there to catch it if you try and re-declare a variable.

$0.02 😁

cormacmccarthy:
Ball is back in your court @interactivellama

js/repeater.js
```diff
@@ -736,40 +737,44 @@
...
+
+						// Extracts subset from object chain (get `items` out of `foo.bar.items`). I think....
+						var propsToRepeatOn = objectAndPropsToRepeatOn.slice(1);
+						for (var prop = 0; prop < propsToRepeatOn.length; prop++) {
+							if (subset[propsToRepeatOn[prop]] !== undefined) {
+								subset = subset[propsToRepeatOn[prop]];
+							} else {
+								subset = [];
+								logWarn('WARNING: Repeater unable to find property to iterate renderItem on.');
+								break;
```
interactivellama: Nice use of `break`. I'm not sure I've seen one in the wild like this recently.

cormacmccarthy: TBH the `break` was in the original code. I took it out for clarity somewhere along the way, and then once the code was simplified to this point I realized it was now actually _less clear_, and so I put it back in. :P

PR #1907, opened 2016-12-08
Datepicker w/ Moment parseDate parsing strings results in good date object not `Invalid Date`

### LINKED ISSUE #1906: Datepicker w/ moment setDate doesn't accept strings
**Reproduce:** When setting datepicker w/ moment via `setDate` pass a string. Then use `getDate`.
**Expecte Results:** a Date object matching the string passed
**Actual Results:** a Date object with `Invalid Date` state

swilliamset:
fixes #1906 

Code made two attempts to parse but passed the failed result from the first attempt to the second attempt instead of the raw date string.


swilliamset:
@futuremint please sanity check now that i moved the dist build before the test.

PR #1921, opened 2017-01-10
chore(package): update grunt-html to version 8.2.0

swilliamset:
debugging travis errors on #1917


swilliamset:
no longer using grunt-html

PR #1161, opened 2015-03-06
Allows for wizard to go RTL

vernak2539:
fixes #1156.

All they have to do is add a class of "rtl" next to wizard. Example of markdown can be found [here](https://gist.github.com/vernak2539/e8e4a9226115a4e042cf)

PR #1930, opened 2017-01-24
[Tree, Repeater]: Fixed some issue in RTL direction

afz:
Hi,
**Trre**
Play icon(glyphicon-play) in both RTL(right to left) and LTR is same shape, also bootstrap not have left direction for glyphicon-play.
In my  opinion using glyphicon-triangle-right instead glyphicon-play for tree is a right choise, because this icon have right/left/up/down direction in bootstrap.
**Repeater**
Fixed action and frozen columns position in RTL direction

### DIFF
--- a/index.html
+++ b/index.html
@@ -2121,7 +2121,7 @@
 								<ul class="tree" id="myPickerTree1" role="tree">
 									<li class="tree-branch hidden" data-template="treebranch" role="treeitem" aria-expanded="false">
 										<div class="tree-branch-header">
-											<button type="button" class="glyphicon icon-caret glyphicon-play"><span class="sr-only">Open</span>
+											<button type="button" class="glyphicon icon-caret glyphicon-triangle-right"><span class="sr-only">Open</span>
 											</button>
 											<button type="button" class="tree-branch-name">
 												<span class="glyphicon icon-folder glyphicon-folder-close"></span>
--- a/js/repeater-list.js
+++ b/js/repeater-list.js
@@ -224,11 +224,11 @@
 			var scrollLeft = $wrapper.scrollLeft();
 			var frozenEnabled = this.viewOptions.list_frozenColumns || this.viewOptions.list_selectable === 'multi';
 			var actionsEnabled = this.viewOptions.list_actions;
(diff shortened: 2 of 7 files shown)


afz:
Added new option to repeater-list for set bidi-direction.
name: list_direction
type: string
default: 'ltr'
values: either 'ltr' or 'rtl'

swilliamset:
Thanks for the updates. On further review can we adjust the approach to be similar to that of the wizard control introduce with #1161. You'll notice a `rtl` class was added to `.wizard` container to drive any necessary changes. This is more consistent with the declarative nature of fuelux.

Please provide a screenshot and codepen of your expected results 

For verification it appears what you are attempting to do is...
- tree - point the arrow to the left
- repeater - swap the select and action columns in repeater 

Also because this is regarding two different controls there should be two PRs.

swilliamset:
@vernak2539 was the author of #1161. i'd like his feedback as well.

vernak2539:
Hey @afz, thanks for the PR. I have a few items on feedback.

1. Don't mind the change of the icon due to the situation as you explained, but I would argue `.glyphicon-chevron-right` and `.glyphicon-chevron-left` are better choices
   * Or if you keep the play icon, you can just use a CSS transform to rotate it
2. I think this entire PR (minus HTML changes) could be accomplished purely as a CSS change (please see PR #1161). Creating a `.rtl` class would allow someone to just add or remove the class and not have to re-instantiate the control to update the text direction. 
3. In your CSS/LESS you will have to add `direction: rtl;`, if we truly want this to be easy for other to implement

afz:
Hi,
about .glyphicon-chevron-left/right, is also OK.
But about repeater-list(action and frozen), when repeater horizontal scrolled, you changed left/right in js file, so there is no way via css to fix this for RTL .

vernak2539:
@afz you're correct. It looks good to me then aside from the fact that you'll need to actually add the CSS class that changes the text direction, which I've put in my comment above.

swilliamset:
@afz when you are able to implement @vernak2539 class suggestion please let us know and we'll reopen.

PR #1941, opened 2017-03-07
serve-static module dependency and references removed

swilliamset:
#1939 and #1929 can be closed after this is merged


futuremint:
This PR needs more explanation around the the "why".

interactivellama:
This seems like it would kill: http://fuelux-edge.herokuapp.com/ . Are we trying to remove that app's deployment?

swilliamset:
the "why" was it looked like we weren't using it. but based @interactivellama's note we are. although when i pinged the url the heroku dyno had to warm up so we aren't using it very often.

Issue #1944, opened 2017-03-16
Repeater sets height to 0 when resized inside of display:none div


austin-rauschuber:
Whenever the repeater is placed inside of another DOM element whose display is set to none, the resize function of the repeater results in the height being set to 0. This ultimately leads to the following issue occurring frequently:

-  Load tab/modal that includes the repeater
-  Close tab/modal
-  Resize screen
-  Reopen tab/modal
Expected Result: Repeater height is resized properly to size of window
Actual Result: Repeater height is set to 0

While you can simply trigger a resize event whenever you open the tab/modal again, this seems unreasonable and "hackish" to include in every location in which the repeater is used.

My suggested solution is to adjust all parents of the repeater that are hidden to be set to "display:block" for the duration of the resize function and then reset to their prior values afterwards. Due to the single-threaded nature of javascript, I believe that the UI thread won't ever be picked up during the execution of the function causing the changes to not affect the browser's rendering of the page.

swilliamset:
closed by #1946

PR #1946, opened 2017-03-17
Set height properly on repeater resize

austin-rauschuber:
Fixes Issue #1944 
This fix resolves the issue of the fuel repeater incorrectly calculating its height whenever the resize function is called when the repeater is not visible, specifically when inside of a parent element which is not visible. The fix relies upon the single threaded nature of javascript to temporarily show the repeater, calculate its height, and then hide the repeater all before the UI thread re-obtains control to display the changes. 

Alternative solutions such as making the items position: absolute and display:hidden were also considered but ultimately resulted in inaccurate height calculations in certain circumstances with no real benefit, since the UI thread does not display the changes regardless.

### DIFF
--- a/js/repeater.js
+++ b/js/repeater.js
@@ -698,6 +698,23 @@
 			var viewTypeObj = {};
 			var height;
 			var viewportMargins;
+			var scrubbedElements = [];
+			var previousProperties = [];
+			var $hiddenElements = this.$element.parentsUntil(':visible').addBack();
+			var currentHiddenElement;
+			var currentElementIndex = 0;
+
+			// Set parents to 'display:block' until repeater is visible again
+			while (currentElementIndex < $hiddenElements.length && this.$element.is(':hidden')) {
+				currentHiddenElement = $hiddenElements[currentElementIndex];
+				// Only set display property on elements that are explicitly hidden (i.e. do not inherit it from their parent)
+				if ($(currentHiddenElement).is(':hidden')) {
+					previousProperties.push(currentHiddenElement.style['display']);
+					currentHiddenElement.style['display'] = 'block';
+					scrubbedElements.push(currentHiddenElement);
+				}
+				currentElementIndex++;
+			}
 
 			if (this.viewType) {
 				viewTypeObj = $.fn.repeater.viewTypes[this.viewType] || {};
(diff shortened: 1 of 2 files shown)


js/repeater.js
```diff
@@ -699,6 +699,13 @@
 			var height;
 			var viewportMargins;
 
+			var $hiddenParents = this.$element.parents().addBack().not(':visible');
```
swilliamset: change to `$hiddenParents = this.$element.parents(':hidden').addBack();`
easier to read

swilliamset: check to see if `$element` is hidden before scanning all it's parents

swilliamset: also limit the set to first `:visible` ancestor. this will minimize the number of parents visited.

js/repeater.js
```diff
@@ -698,6 +698,21 @@
 			var viewTypeObj = {};
 			var height;
 			var viewportMargins;
+			var $hiddenParents = [];
+			var previousProperties = [];
+			var $previousParents = this.$element.parentsUntil(':visible').addBack();
```
swilliamset: instead of `$previousParents` please call it `$hiddenElements`

js/repeater.js
```diff
@@ -698,6 +698,21 @@
...
+			var currentParent;
+			var currentParentIndex = 0;
+
+			while (currentParentIndex < $previousParents.length && this.$element.is(':hidden')) {
```
swilliamset: `$this.$element` check is not necessary because of hidden parent check above

js/repeater.js
```diff
@@ -698,6 +698,21 @@
 			var viewTypeObj = {};
 			var height;
 			var viewportMargins;
+			var $hiddenParents = [];
+			var previousProperties = [];
+			var $previousParents = this.$element.parentsUntil(':visible').addBack();
+			var currentParent;
```
swilliamset: looking at the repeater element as well so this variable name could be `$currentHiddenElement`

js/repeater.js
```diff
@@ -698,6 +698,21 @@
...
+			var currentParentIndex = 0;
+
+			while (currentParentIndex < $previousParents.length && this.$element.is(':hidden')) {
+				currentParent = $previousParents[currentParentIndex];
+				if ($(currentParent).is(':hidden')) {
```
swilliamset: is this necessary? we are now only looping through hidden elements correct?

PR #1959, opened 2017-04-14
add validation support to templates

swilliamset:
Uses the following attributes in templates

*  _new_ `validation` [validation state string](http://getbootstrap.com/css/#forms-control-validation)
* `required` boolean

Adds `.required` to `.form-group` for asterisk by `.control-label`

### DIFF
--- a/index.html
+++ b/index.html
@@ -58,7 +58,7 @@
 				<h2>Checkboxes</h2>
 				<div class="thin-box">
 					<section id="checkboxes-block">
-
+						<div id="myCheckbox1Wrapper" style="width: 300px;"></div>
 						<div class="checkbox" id="myCheckbox">
 							<label class="checkbox-custom" data-initialize="checkbox" id="myCustomCheckbox1">
 								<input class="sr-only" type="checkbox" value="" />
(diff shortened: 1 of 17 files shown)


cormacmccarthy:
What markup did you use to test? The index.html page has not been updated to use hbs or underscore templates yet, and the markup on that page is incompatible with this change.

cormacmccarthy:
Please update the `markup/` folder contents to reflect these changes.

The templates are not covered by the unit tests currently so there is no way for me to test these changes in Fuel UX 3 without writing a bunch of code. We have four options here:

1. refactor index.html to use the templates instead of hard-coded markup so this can be manually visually tested (medium-small effort)
2. refactor all affected unit tests to import templates and match their output against markup/ folder contents (medium-large effort)
3. Test manually/visually using fuelux site locally (and hope that the site isn't doing anything that would make it work there but not in Fuel UX proper)
4. Do 1 & 2

IMO, the "correct" thing to do would be 4, as it aligns with our "Engineering Thinking" value and our "trust" and "automation" methods. I'd recommend breaking the work up into individual PRs, one for each component, so that we have steady demonstrable deliverables and if a higher priority task comes up already completed work won't get lost or delayed. 

I'd be happy to help tackle this with you.

PR #1960, opened 2017-04-17
Make spinbox trigger changed after setValue

### LINKED ISSUE #1950: Using Spinbox setValue leads to inconsistent results
Hi,

I initialize a spinbox with startvalue 0. 
Next I call $(...).spinbox('setValue', 10), and the spinbox value changes to 10.
If i now type 0 into the spinbox, no change event is fired.

The reason is likely that neither setValue(val) nor value(val) functions to affect this.lastValue, which is used for comparison with the new value before triggering the change event.

vargasd:
After programmatically setting the value of the spinbox, the previous
value would be retained in the lastValue property of the control,
causing the "changed" event not to get triggered if the user were to
enter the previous value again. To avoid this, the current setValue
method was converted into a private method with an additional parameter
to set the last value, and the public method will wrap this and pass
that parameter, while internal uses of the method will use the private
method to avoid other side effects.

Fixes #1950

### DIFF
--- a/test/spinbox-test.js
+++ b/test/spinbox-test.js
@@ -285,6 +285,21 @@ define( function ( require ) {
 
 	} );
 
+	QUnit.test('spinbox should trigger change after using setValue', function (assert) {
+		var $spinbox = $(html).find('#MySpinboxDecimal').spinbox({
+			value: '1'
+		});
+
+		$spinbox.spinbox('setValue', '2');
+
+		$spinbox.on('changed.fu.spinbox', function () {
+			assert.ok(true, 'spinbox triggers changed after input' );
+		});
+
+		$spinbox.find('.spinbox-input').val(1);
+		$spinbox.find('.spinbox-input').focusout();
+	});
+
 	QUnit.test( "should destroy control", function( assert ) {
 		var $el = $( html ).find( "#MySpinbox" );
 
--- a/js/spinbox.js
+++ b/js/spinbox.js
@@ -167,11 +167,11 @@
 		},
 
 		render: function render() {
-			this.setValue(this.getDisplayValue());
+			this._setValue(this.getDisplayValue());
 		},
 
 		change: function change() {
-			this.setValue(this.getDisplayValue());
+			this._setValue(this.getDisplayValue());
 
 			this.triggerChangedEvent();
 		},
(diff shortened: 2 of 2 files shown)


futuremint:
@vargasd please look into the failing unit test and ensure they are all passing.

vargasd:
Should I open another PR to fix the failing unit tests? They were previously broken and don't relate to this issue.

test/spinbox-test.js
```diff
@@ -285,6 +285,21 @@ define( function ( require ) {
...
 
+	QUnit.test('spinbox should trigger change after using setValue', function (assert) {
+		var $spinbox = $(html).find('#MySpinboxDecimal').spinbox({
+			value: '1'
+		});
+
+		$spinbox.spinbox('setValue', '2');
+
+		$spinbox.on('changed.fu.spinbox', function () {
+			assert.ok(true, 'spinbox triggers changed after input' );
```
futuremint: This is probably not even getting validated.  You need to use `assert.async` in here and call it to ensure this assertion gets asserted. This is an async test, so you need to tell QUnit it is async.

vargasd: This isn't asynchronous. The handler should be called synchronously on the focusout. I can make it asynchronous, but then the failure would happen because of a timeout rather than a 0-assertion test. I prefer the latter because it means the tests run more quickly, but I can change it if you want.

futuremint:
I didn't know the unit tests were previous broken.  You don't need to open a PR, @swilliamset can look into those.  I just wanted to double check they weren't from your code changes is all.

swilliamset:
unit tests are failing on repeater tests the same as #1959. i'm looking into it.

vargasd:
Hey @swilliamset, I have rebased and tests are now passing.

scotthetrick:
@swilliamset Now that Sam has rebased and the tests are passing, is there anything else you need to complete the review?

swilliamset:
i'll take a look at this on Monday.

scotthetrick:
@swilliamset any update on this review?  Thanks.

swilliamset:
My apologies for the delay. This is priority 0 today.

test/spinbox-test.js
swilliamset: this is fine for now but for future reference when using an event listener an async test is more appropriately because it is unknown what is happening before the event is triggered.

PR #1974, opened 2017-05-18
Clean repeater list extension row content of XSS (fix #1973)

### LINKED ISSUE #1973: Repeater row content XSS
Include script tag in data passed to rows of the repeater. The script will execute.

```
<script>alert('attack!');<script>
```

will result in an alert message

i would provide a codepen but it has problems with the script tag

swilliamset:
fix #1973 

Added unit test structure for multiple known XSS patterns and ease of adding more.

Updated existing clean method to prevent double encoding.

### DIFF
--- a/js/repeater-list.js
+++ b/js/repeater-list.js
@@ -16,18 +16,23 @@
 (function UMDFactory (factory) {
 	if (typeof define === 'function' && define.amd) {
 		// if AMD loader is available, register as an anonymous module.
-		define(['jquery', 'fuelux/repeater', 'fuelux/checkbox'], factory);
+		define(['jquery', 'fuelux/utilities', 'fuelux/repeater', 'fuelux/checkbox'], factory);
 	} else if (typeof exports === 'object') {
 		// Node/CommonJS
-		module.exports = factory(require('jquery'), require('./repeater'), require('./checkbox'));
+		module.exports = factory(require('jquery'), require('./utilities'), require('./repeater'), require('./checkbox'));
 	} else {
 		// OR use browser globals if AMD is not present
 		factory(jQuery);
 	}
 }(function repeaterList ($) {
+	if (!$.fn.utilities) {
+		throw new Error('Fuel UX repeater control list extension requires FuelUX utilities.');
+	}
 	// -- END UMD WRAPPER PREFACE --
 
 	// -- BEGIN MODULE CODE HERE --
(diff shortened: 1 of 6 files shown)


swilliamset:
@TheJaredWilcurt please review this as well

swilliamset:
After review with @TheJaredWilcurt this needs further discussion. 

Between customizing the content of data returned by the [`dataSource` `data` object](http://getfuelux.com/extensions.html#bundled-extensions-thumbnail-datasource) or the usage of [`list_columnRendered` option](http://getfuelux.com/extensions.html#bundled-extensions-list-options) the responsibility of ensuring the data is xss clean is up to the consuming code. 

The current approach would be a breaking change. Any markup returned by the `dataSource` will be encoded and not render. Instead applications would need to sure only straight text is passed via the `dataSource` and any custom markup (ie a link) should be provided using `list_customRendered`.

To not be a breaking change `cleanUp` could include an option for a less strict regex that only matches `<script>` or agreed upon pattern to recognize unsafe code. This can be super complicated.

Possible options:
* reject solution entirely - all xss responsibility on consuming code
* keep solution - document only pure text (no markup) should be returned from `dataSource`
* modified solution - only encode unsafe markup returned by `dataSource`

cc @futuremint @cormacmccarthy

swilliamset:
Closing as `wontfix`. 

XSS free data is the responsibility of the consuming application. It is recommended to use the `dataSource` object to insure data is xss free. The `cleanInput` utility used in this solution could be used to scrub the data. 

In the spirit of separation of concerns any markup used to enhance the data (links, icons, etc) should be introduced via the `list_columnRendered` hook. This will insure `dataSource` data is markup free.

cormacmccarthy:
As we discussed yesterday, I don't believe that dataSource actually should be markup free. I believe that dataSource is where the pre-render actions take place. `list_columnRendered` is only for things that must take place after something has been rendered to the page (like checking to see if something is visible, or measuring it). Any formatting or html templating should take place prior to attaching to the DOM because of the massive performance hit that doing those things in `list_columnRendered` (after they have been attached to the DOM) can be. If you have 100 rows, each with 10 columns, then running `list_columnRendered` 1000 times and doing all of your DOM manipulation there on "data" that has already been attached can and has caused page crashes.

The misnamed, so-called, "dataSource" is really more of a "contentSource". It both gets the data from the server (or itself) and takes care of all of the rendering.

PR #1980, opened 2017-05-22
refactor `cleanInput` utility method  to prevent double encoding

swilliamset:
instead of looking for markup and sanitizing this approach looks for encoding, decodes it, and re encodes everything via jquery.


swilliamset:
[9b12717](https://github.com/ExactTarget/fuelux/pull/1980/commits/9b127176c44f5842611937be8f0d9c9139bae529) I hate adding dependencies explicitly for an ide but if i recall there have been other incidence of ide's not integrating with our rules so thought this would help

PR #1984, opened 2017-05-31
Makes tree accessible

### LINKED ISSUE #1964: Tree should be accessible
As part of the a11y review of one of our apps, a tree nodes can be tabbed through, but a true tree view uses arrow keys. 

- [x] Links within nodes and expand buttons have an implicit `tabindex=0` currently (this is incorrect)
- [x] All nodes of tree (including expand/collapse buttons on folders with children) should have `tabindex=-1`. 
- [x] Root/parent of tree should have `tabindex=0`
- [x] The root element of the tree should always contain an attribute `aria-activedescendant`[1] that is set to the id of the leaf/branch that has focus.
- [x] Create issue in fuelux-site repo to document aria requirements for tree, specifically sub-nodes (leaves/branches) because developers are able to insert custom html in branches and they need to set those element's `tabindex`s to `-1` (see [SLDS](https://latest-208.lightningdesignsystem.com/components/lookups) for an example of documenting accessibility --although we don't need to be as thorough since we control most of the JS etc).

Expected behavior for a tree according to [w3c](http://w3c.github.io/aria-practices/#TreeView)
- [x] When a single-select tree receives focus:
    - [x] If none of the nodes are selected before the tree receives focus, focus is set on the first node.
    - [x] If a node is selected before the tree receives focus, focus is set on the selected node.
- [x] When a multi-select tree receives focus:
    - [x] If none of the nodes are selected before the tree receives focus, focus is set on the first node.
    - [x] If one or more nodes are selected before the tree receives focus, focus is set on the first selected node.
- [x] Right arrow:
    - [x] When focus is on a closed node, opens the node; focus does not move.
    - [x] When focus is on a open node, moves focus to the first child node.
    - [x] When focus is on an end node, does nothing.
- [x] Left arrow:
    - [x] When focus is on an open node, closes the node.
    - [x] When focus is on a child node that is also either an end node or a closed node, moves focus to its parent node.
    - [x] When focus is on a root node that is also either an end node or a closed node, does nothing.
- [x] Down Arrow: Moves focus to the next node that is focusable without opening or closing a node.
- [x] Up Arrow: Moves focus to the previous node that is focusable without opening or closing a node.
- [x] Home: Moves focus to the first node in the tree without opening or closing a node.
- [x] End: Moves focus to the last node in the tree that is focusable without opening a node.
- [x] Enter: activates a node, i.e., performs its default action. For parent nodes, one possible default action is to open or close the node. In single-select trees where selection does not follow focus (see note below), the default action is typically to select the focused node.
    - [x] If folder
        - [x] If selectable, select folder
        - [x] If not selectable toggle folder
    - [x] If item, select item
    - [x] If “load more”, click it and set focus on first element loaded.

Stretch Goals:
~- [ ] Type-ahead is recommended for all trees, especially for trees with more than 7 root nodes:
    - [ ] Type a character: focus moves to the next node with a name that starts with the typed character.
    - [ ] Type multiple characters in rapid succession: focus moves to the next node with a name that starts with the string of characters typed.~

1: `aria-activedescendant` is used when a composite widget is responsible for managing its current active child to reduce the overhead of having all children be focusable. Examples include: multi-level lists, trees, and grids. In some implementations the user agent may use aria-activedescendant to tell assistive technologies that the active descendant has focus. 

Additional a11y links for reference (however as with most a11y specs, I'm not sure all the information is accurate):
- https://www.w3.org/WAI/GL/wiki/Using_ARIA_trees
- http://www.oaa-accessibility.org/examplep/treeview1/
- http://w3c.github.io/aria-practices/#TreeView
- https://www.w3.org/WAI/GL/wiki/Using_aria-activedescendant_to_allow_changes_in_focus_within_widgets_to_be_communicated_to_Assistive_Technology
- https://www.marcozehe.de/2012/02/06/if-you-use-the-wai-aria-role-application-please-do-so-wisely/

@arodenbeck should review when complete

cormacmccarthy:
fixes #1964


grunt/tasks/test.js
```diff
@@ -3,6 +3,10 @@ module.exports = function test (grunt) {
 	grunt.registerTask('test', 'run jshint, qunit source w/ coverage, and validate HTML',
 		['jshint', 'connect:testServer', 'qunit:noMoment', 'qunit:globals', 'test-dist', 'htmllint']);
 
+	grunt.registerTask('unittest', 'run jshint, qunit source w/ coverage, and validate HTML',
```
swilliamset: why i don't see `unittest` being used anywhere

js/tree.js
```diff
@@ -32,46 +34,57 @@
 	// TREE CONSTRUCTOR AND PROTOTYPE
 
 	var Tree = function Tree(element, options) {
-		this.$element = $(element);
+		var $element = $(element);
```
swilliamset: I only see `$element` used once at https://github.com/ExactTarget/fuelux/pull/1984/files?#diff-e5768895bd9d9f525527077130cf278bR70

It could be replaced with 
```
this.$element.on('focus', function setFocusOnTab () {
	var $tree = $(this);
	focusIn($tree, $tree);
});
```

making this assignment unnecessary.

interactivellama:
- Needs visual focus indicator SLDS currently has:
```
.slds-tree__item a:focus {
  outline: 0;
  text-decoration: underline; }
```
- Shift+tab isn't working (might need less of a keyboard trap)
- `ul role="tree` node needs either `aria-label` or `aria-labelledby`
- selected items need `aria-selected`
- tree needs `aria-multiselectable` unless it is single selection. multi should be default
- if tree is single select, `aria-multiselectable` needs to be false. We've handled this in the past through the dataSource I think. I'm not sure how to expose with the API.

cormacmccarthy:
Working through @interactivellama's feedback now.

I believe we agreed not to do this:

- Shift+tab isn't working (might need less of a keyboard trap). This should just work based on tabindex.

Doing these:

- [ ] Needs visual focus indicator SLDS currently has
- [ ] ul role="tree" node needs either aria-label or aria-labelledby
- [ ] selected items need aria-selected="true"
- [ ] tree needs aria-multiselectable unless it is single selection. multi should be default.
- [ ] if tree is single select, aria-multiselectable needs to be false

cormacmccarthy:
Note that the tree markup changes slightly with this PR adding aria-* attributes and tabindex to many of the nodes in tree. It's a non-breaking change.

PR #1990, opened 2017-06-08
Adds visual regression testing to Fuel UX

### LINKED ISSUE #1986: Fuel UX Tests should include visual regression testing
p'raps [Niffy](https://segment.com/blog/perceptual-diffing-with-niffy/)?

cormacmccarthy:
fixes #1986 

The regression tests require an express server to be spun up and render handlebars templates on the server side so that what is served to the page is static HTML. This is because Fuel UX is declarative and some components either initialize on page load or on user interaction (eg. checkbox). In order to accomplish this, the component has to be rendered and ready on pageload when Fuel UX scrapes the page looking for components to initialize.

The Fuel UX Handlebars templates are used to accomplish this.

The test/regression folder should have a separate file for each Fuel UX asset.

The regression tests will look at either the local dev code (from `js/*` or `dist/css/fuelux.css`) or will look at the reference code in `reference/dist/`. Ideally the reference code will match the last prod release. Upon Fuel UX release the code from `/dist/` is copied into `/reference/dist/`. We use the dist version of CSS so that we don't have to deal with LESS crap in the browser.

If a change intends to modify the look of an asset, those changes should be manually copied into the reference directory. The reason this was done was that otherwise we would have to "ignore" failing tests when you made a change that modified the look and feel. Now since you will be manually copying the changes into the reference folder, you are intentionally changing the output of reference and when you run the tests they will merely be making sure they match. Further changes tested against this new output will make sure that the change did not revert.


interactivellama:
In Jest snapshot testing which is kind of similar, but strings of DOM or HTML, there's an overwrite command `jest snapshot -- -u` or something like that that basically says, all this stuff is correct--I'm 100% sure--please overwrite all the references. Then git only notices the ones that change and we version that.

Is there something like that that can be done?

cormacmccarthy:
done.

cormacmccarthy:
Of _course_ it doesn't work on Travis 😐

futuremint:
Looks like niffy hardcodes the filepath to `/tmp/niffy/`: https://github.com/segmentio/niffy/blob/master/index.js#L186

test/REGRESSION_TESTING.md
```diff
@@ -0,0 +1,21 @@
...
+
+The Fuel UX Handlebars templates are used to accomplish this.
+
+The test/regression folder has a separate file for each Fuel UX asset.
+
+The regression tests will look at either the local dev code (from `js/*` or `dist/css/fuelux.css`) or will look at the reference code in `reference/dist/`. Ideally the reference code will match the last prod release. Upon Fuel UX release the code from `/dist/` is copied into `/reference/dist/`.
+
+If a change intends to change the look of an asset, you will need:
+
+1. Make a build (`npm build`)
```
interactivellama: `npm run build`

cormacmccarthy:
Done! All ready for you again @futuremint

PR #1995, opened 2017-06-20
adds tabindex to checkbox and radio

### LINKED ISSUE #1972: Checkbox and Radio should be accessible
Checkbox and Radio Button templates should include `tabindex="0"` so that in applications that use CSS to hide the input element (eg Lightning Theme) they are still tab-able.

cormacmccarthy:
fixes #1972 

Cleans up all checkbox unit tests as well. Determined afterwards that I didn't actually need to add any unit tests because we don't need to test native browser functionality (that, and I couldn't figure out how to actually even do it. Keyboard testing is... problematic...)


reference/markup/checkbox-inline-no-js.html
```diff
@@ -9,6 +9,7 @@
 		name="checkboxes-new-inline--1"
 		checked="checked"
 		value=""
+		tabindex="0"
```
interactivellama: Is this needed?

cormacmccarthy: It is if someone set's visibility to "hidden". It also isn't _hurting_ anything, so even if it isn't "needed" it's better safe than sorry. If the implementor knows enough to make the call to not put it in their code, that's fine. I guess it's like buckling your seat-belt. 99% of the time, it's a waste of time...

cormacmccarthy:
Nope. This isn't doing what I thought it was doing. `tabindex="0"` doesn't fix the issue.

PR #1996, opened 2017-06-21
Ch1972   add accessibility check for checkbox and radio

### LINKED ISSUE #1972: Checkbox and Radio should be accessible

cormacmccarthy:
Fixes #1972 

Adds a console error message if `visibility: hidden` or `visibility: collapse` is applied to the checkbox.

Cleans up checkbox unit tests and makes UMD wrapper more consistent across all core fuel ux files.

If you are getting the following error:
`For accessibility reasons, in order for tab and space to function on [checkbox|radio], `visibility` must not be set to `hidden` or `collapse`. See https://github.com/ExactTarget/fuelux/pull/1996 for more details.`

It means that your code (probably in the CSS) is inappropriately setting the checkbox or radio `visibility` to `hidden` or `collapse`. You can fix it by:

1. (Preferred) Update to the most recent version of Lightning Theme or MC Theme.
2. (Allowed) If you are unable to upgrade to the latest Lightning or MC Themes, or if you are using some other custom CSS or theme that sets this, find the place where your code is doing this and don't do it instead.
    - For Lightning Theme, that means editing [file] line [number] to be [thing]
    - For MC Theme, that means editing [file] line [number] to be [thing]
3. If you have a valid reason for making the checkbox or radio hidden (ie, if it is hidden on pageload and you plan to show it later) try making the parent container hidden instead, or else setting `display` to `none` on pageload. Or (worst-case) add a data attribute `data-ignore-visibility-check=true` or pass an option on init `$yourCheckbox.checkbox({ignoreVisibilityCheck: true})` to make checkbox ignore the error.


interactivellama:
I really like this proactive check. You might consider some kind of production/dev env check, but I think you can only do that in CommonJS not ES5.

index.js
```diff
@@ -10,9 +10,9 @@ define(function (require) {
 			window.console.log.apply(console, args);
 		}
 	};
-	if (window.console && window.console.clear) {
-		window.console.clear();
-	}
+	// if (window.console && window.console.clear) {
```
interactivellama: Just remove it, if you don't want it.

js/scheduler.js
```diff
@@ -23,14 +25,13 @@
...
-}(function ($) {
-	if (!$.fn.combobox || !$.fn.datepicker || !$.fn.radio || !$.fn.selectlist || !$.fn.spinbox) {
-		throw new Error('Fuel UX scheduler control requires combobox, datepicker, radio, selectlist, and spinbox.');
-	}
-
+}(function SchedulerWrapper ($) {
 	// -- END UMD WRAPPER PREFACE --
 
 	// -- BEGIN MODULE CODE HERE --
+	if (!$.fn.combobox || !$.fn.datepicker || !$.fn.radio || !$.fn.selectlist || !$.fn.spinbox) {
```
interactivellama: I don't think any of these namespace checks should be in the bundled code, since the code is known to be there. You've moved it into the bundled library in multiple place.

cormacmccarthy: I literally just moved it down below the UMD wrapper comment... Are you saying that somehow changed the way it gets bundled?

interactivellama: I'm saying that if you move it into the `-- BEGIN MODULE CODE HERE --` part then it WILL get into the bundle, instead of just used when you require the one control.

test/checkbox-test.js
```diff
@@ -1,315 +1,33 @@
...
+			assert.ok(this.$checkedEnabled.checkbox() === this.$checkedEnabled, 'checkbox should be initialized');
+		});
+
+		require('./checkbox-tests/initialization')(QUnit);
+		require('./checkbox-tests/disable')(QUnit);
+		require('./checkbox-tests/checked')(QUnit);
+		require('./checkbox-tests/return')(QUnit);
+		require('./checkbox-tests/get-value')(QUnit);
+		require('./checkbox-tests/events')(QUnit);
+		require('./checkbox-tests/visibility')(QUnit);
```
interactivellama: if we are skipping the visibility tests, then should we comment this line out so no one sees it?

interactivellama:
`index.html` should be an example of how to use Fuel UX, not a testing location for console errors. Please move these to another page if in browser testing is the purpose.

js/scheduler.js
cormacmccarthy: Wait, so you're telling me that the bundle gets generated in part based on a comment in the code?

test/checkbox-test.js
cormacmccarthy: Well, the point of making it `.skip` is that you alert the person that "there's a thing that maybe ought to be tested, but actually isn't". That way they can either add the test, or go test it themselves. If we comment it out, we might as well just remove it... Any thoughts @futuremint (Dave is the one who told me to add the skip tests)?

interactivellama:
Please remove skipped tests or add something obvious to tell new folks to the project why they are skipped. Could be in test name?

screen shot 2017-06-27 at 3 36 39 pm

js/scheduler.js
interactivellama: Yes, the bundle is created with concat.

https://github.com/ExactTarget/fuelux/blob/master/grunt/config/concat.js#L32

test/checkbox-test.js
interactivellama: How about something like "skipped intentionally, future work" added to the test name.

cormacmccarthy: Looking closer at this, checkbox visibility tests are fully defined. Not sure where you are seeing it is skipped? Tree is the only thing that has skipped tests (and those were added by a different PR).

cormacmccarthy:
The skipped tests were added in a different PR, but I've documented them in this PR a little in response to your request for them to be removed. In #2000 I fleshed a few of them out more. 

It's important for them to be there as an indicator that there is functionality that isn't actually being tested (that in a perfect world would be). That way, if the PR or change someone is working on has anything to do with those things (this PR doesn't...), they'll know that those things need tested more carefully manually by both the dev and the reviewer(s) and/or those tests need fleshed out.

cormacmccarthy:
Oops! I always forget that index.html is anything other than just a place to do testing. Will fix, thanks.

PR #1998, opened 2017-06-23
adds initialized.fu.tree event. Adds static datasource consumer to tr…

cormacmccarthy:
…ee so people can just pass in a JSON object representation of their entire tree


interactivellama:
Please remove skipped tests or add something obvious to tell new folks to the project why they are skipped. Could be in test name?

screen shot 2017-06-27 at 4 20 02 pm

PR #2000, opened 2017-06-24
Gh1994   keyboard navigation of open empty folder in tree should work as expected

### LINKED ISSUE #1994: Keyboard navigation of open empty folder in Tree
Pressing down arrow key on an open empty folder does not focus on next folder. I think it's looking for children.

cormacmccarthy:
fixes #1994

~blocked by #1996 #1998~

Makes tree function as expected when navigating over open empty branches. Adds unit tests.


interactivellama:
This doesn't seem to fix the empty tree node problem.

Go to `http://localhost:8000/dev/staticDataTree.html` Open the first one and then try to navigate it with the keyboard.

cormacmccarthy:
Odd. What browser are you in? Works for me.

index.html
```diff
@@ -59,14 +59,14 @@
 				<div class="thin-box">
 					<section id="checkboxes-toggling">
 						<div class="checkbox">
-							<label class="checkbox-custom" data-initialize="checkbox" id="myCustomTogglingCheckbox1">
```
interactivellama: These are being added back in.

interactivellama:
empty branch tree stuff is good. I was in the wrong branch, when I was testing before 😈 . Please clean up changes from other branches.

Please add links to `http://localhost:8000/dev/staticDataTree.html` on `index.html` in the tree section, so folks don't have to search for them.

Issue #2011, opened 2017-09-08
Keydown Event Listener Stops Propagation of all Events in Tree


jacobroschen:
Steps to reproduce:
1. Add a child element to the tree, for example a popover with an input box for renaming
2. Type in a new name with spaces
3. Observe that the space button does **not** work

Space, enter, up, down, left, right, home, and end are all consumed because they are being used for the new keyboard shortcuts in 3.16.1. I expect keyboard shortcuts to still work, and for the event listener to only listen on the list items, not all child elements in the tree.

Expected results:

1. Add a child element to the tree, for example a popover with an input box for renaming
2. Type in a new name with spaces
3. Observe that the space button **does** work

PR #2012, opened 2017-09-08
Fix keydown event listener scope in tree

jacobroschen:
fixes https://github.com/ExactTarget/fuelux/issues/2011

This adds a unit test demonstrates what happens when a child element is added along with the fix, which limits the scope of the keydown event consumption.

### DIFF
--- a/js/tree.js
+++ b/js/tree.js
@@ -71,6 +71,13 @@
 		});
 
 		this.$element.on('keydown', function processKeypress (e) {
+			var target = $(e.target);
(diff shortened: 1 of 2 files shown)


bohawi:
The problem we are trying to solve here is that the accessibility changes are calling preventDefault when using the spacebar in an input in the tree control.

bohawi:
@swilliamset @cormacmccarthy @interactivellama would any of you be available to give this PR a look?  If all is good would you be able to help us get it tagged for 210?

PR #2039, opened 2018-04-17
Prevent infinite scroll from fetching all pages

jacobroschen:
Infinite scroll could sometimes fetch all pages on initialization if the
repeater was not attached to the DOM.


bohawi (approved):
There is still some risk if any of these values are not defined, but if we haven't run into it yet it isn't likely to occur.

PR #2049, opened 2018-07-30
W-5234018 - Fix XSS vulnerability in email admin screen

jbennie44:
This change fixes a P1 security ticket (W-5234018). The issue is that XSS is firing in the DE selector tree in the From Address Management page. The best solution to this problem is to render the name of the the DEs as text rather than as html. As a result, this change needs to be made to change one .html call to a .text call in tree.js.

Here is a link to the P1 sec ticket for reference:
https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000005FEeWIAW/view

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,0 +1,3 @@
+{
+  "registry": "https://registry.bower.io"
+}
--- a/js/tree.js
+++ b/js/tree.js
@@ -159,7 +159,7 @@
 						.removeClass('hide hidden')// jQuery deprecated hide in 3.0. Use hidden instead. Leaving hide here to support previous markup
 						.removeData('template')
 						.removeAttr('data-template');
-					$entity.find('.tree-' + nodeType + '-name > .tree-label').text(treeNode.text || treeNode.name);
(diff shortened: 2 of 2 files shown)


bohawi:
@interactivellama @cormacmccarthy As old-school maintainers of FuelUX, do you guys know if was intended to allow people to use HTML as a tree label?  I don't see anywhere in the documentation where it says that you can't.

jbennie44:
@interactivellama Do you think making this change, even to fix a P1 security ticket, is too risky then? I don't want to break others that are using this... but at the same time I'm curious if supporting this in it's current state is a good thing to do from a security perspective in general.

interactivellama:
I don't own this project anymore, but I like adding the option. You could go a step farther and detect if HTML is in the string, and do a console warning in dev mode (process.env.NODE_ENV !== 'production') although that's not a paradigm used anywhere else in this library, so I don't think that is an easy option.

I noticed there aren't any release notes on the [last four releases](https://github.com/ExactTarget/fuelux/releases). This is something you'd want to put there.

bohawi:
CI reports this error: Error: 0.001% different, open /tmp/niffy/component/repeater-single/diff.png

I pulled down the repo and ran the tests locally and they all passed.

Issue #2064, opened 2019-03-22
bootstrap-psirt


infolock:
**Bootstrap Vulnerabilities**

Primary Source: [https://snyk.io/vuln/npm:bootstrap](https://snyk.io/vuln/npm:bootstrap)

There are a handful of bootstrap components which have XSS vulnerabilities.   These are almost all due to attribute values not being properly sanitized before being used as a selector (which can cause the scripts/html to be executed).

For each of the below, teams that are already on the 3.3.x series of bootstrap can simply upgrade to 3.4.1+ to fix this.

With that issue being reported, we would need to update Fuel3 in general since it relies on Bootstrap 3.x.

More information along with a list of vulnerabilities can be found on the [snyk website](https://snyk.io/vuln/npm:bootstrap).

PR #2065, opened 2019-03-22
Issue #2064 - Bootstrap PSIRT

infolock:
**Bootstrap Vulnerabilities**

Primary Source: [https://snyk.io/vuln/npm:bootstrap](https://snyk.io/vuln/npm:bootstrap)

There are a handful of bootstrap components which have XSS vulnerabilities.   These are almost all due to attribute values not being properly sanitized before being used as a selector (which can cause the scripts/html to be executed).

For each of the below, teams that are already on the 3.3.x series of bootstrap can simply upgrade to 3.4.1+ to fix this.

With that issue being reported, we would need to update Fuel3 in general since it relies on Bootstrap 3.x.

More information along with a list of vulnerabilities can be found on the [snyk website](https://snyk.io/vuln/npm:bootstrap).


infolock:
Could one of you please review this?  This is for the bootstrap PSIRT that has been raised.

@bohawi @tweettypography @jacobroschen

bohawi:
A small concern with this is that the bootstrap upgrade will enable sanitization mode by default for tooltips.  If a consumer is using certain HTML in tooltips this could start to break their functionality with the next `bower update` or `npm update`.  We ran into this issue in the Content Builder application which has a tooltip that contains a `<form>`

infolock:
Agreed.   We would just need a way to just inform everyone about this so they are made aware of it and what to expect.   `sanitize: false` would simply be necessary when `html: true` in that case i think

package-lock.json
```diff
@@ -2062,21 +2047,21 @@
       "dev": true
     },
     "eslint": {
-      "version": "4.18.0",
-      "resolved": "https://registry.npmjs.org/eslint/-/eslint-4.18.0.tgz",
-      "integrity": "sha512-Ep2lUbztzXLg0gNUl48I1xvbQFy1QuWyh1C9PSympmln33jwOr8B3QfuEcXpPPE4uSwEzDaWhUxBN0sNQkzrBg==",
+      "version": "4.19.1",
+      "resolved": "https://artifactory-dev.exacttarget.com:443/artifactory/api/npm/sfmc-npm/eslint/-/eslint-4.19.1.tgz?dl=https://registry.npmjs.org/eslint/-/eslint-4.19.1.tgz",
```
bohawi: @infolock Are there any concerns with using the Artifactory URL in here?

infolock: ah - good call.

PR #2066, opened 2019-03-26
Issue #2064: Bootstrap psirt update

infolock:
**Bootstrap Vulnerabilities**

Primary Source: [https://snyk.io/vuln/npm:bootstrap](https://snyk.io/vuln/npm:bootstrap)

There are a handful of bootstrap components which have XSS vulnerabilities.   These are almost all due to attribute values not being properly sanitized before being used as a selector (which can cause the scripts/html to be executed).

For each of the below, teams that are already on the 3.3.x series of bootstrap can simply upgrade to 3.4.1+ to fix this.

With that issue being reported, we would need to update Fuel3 in general since it relies on Bootstrap 3.x.

More information along with a list of vulnerabilities can be found on the [snyk website](https://snyk.io/vuln/npm:bootstrap).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,13 @@
 <!-- TOC depthFrom:1 depthTo:6 withLinks:1 updateOnSave:1 orderedList:0 -->
 
+Archive Statement:
+
+As of March 2019, this repository is read-only as Salesforce has archived the FuelUX open-source UI framework and will no longer be supported.
+
+Salesforce has introduced a new programming model for the Lightning Component Framework called [Lightning Web Components](https://trailhead.salesforce.com/en/content/learn/projects/quick-start-lightning-web-components) which you are encouraged to check out instead. This programming model is designed for building fast components using modern JavaScript and Web Standards.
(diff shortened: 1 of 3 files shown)


infolock:
@bohawi this is the new PR without any changes to `package-lock.json`.  we _could_ include them, but our internal artifactory domain (like you pointed out earlier) shows up.  so this would keep it omitted and would only be updated after `npm install` is run.

To recap previous discussions:

```
@bohawi commented 4 days ago
A small concern with this is that the bootstrap upgrade will enable sanitization mode by default for tooltips. If a consumer is using certain HTML in tooltips this could start to break their functionality with the next bower update or npm update. We ran into this issue in the Content Builder application which has a tooltip that contains a <form>

@infolock commented 4 days ago • 
Agreed. We would just need a way to just inform everyone about this so they are made aware of it and what to expect. sanitize: false would simply be necessary when html: true in that case i think. that or you also have the ability to provide your own sanitizer
```

PR #2076, opened 2020-03-06
Fix reporting of percentage for infinite scroll

0ndre:
This is a fix for the repeater control when used with the thumbnail extension. In some cases the infinite scroll control would fetch extra pages.


garrisaurus:
👍

